PR #10, opened 2025-02-06
Feat/non return cheatcode

odyslam:


src/CredibleTest.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity 0.8.28;
+pragma solidity ^0.8.13;
```
fredo, line 2: why do we downgrade?

fredo (approved):
lgtm

PR #13, opened 2025-02-14
feat: add trigger recorder

0xgregthedev:


src/Assertion.sol
```diff
@@ -2,10 +2,18 @@
...
 import {Credible} from "./Credible.sol";
+import {TriggerRecorder} from "./TriggerRecorder.sol";
 
 /// @notice Assertion interface for the PhEvm precompile
 abstract contract Assertion is Credible {
-    /// @notice Returns all the fn selectors for the assertion contract
-    /// @return An array of bytes4 selectors
-    function fnSelectors() external pure virtual returns (bytes4[] memory);
+    //Trigger recorder address
+    TriggerRecorder constant triggerRecorder = TriggerRecorder(address(uint160(uint256(keccak256("TriggerRecorder")))));
```
odyslam, line 10: Somewhat irrelevant to this PR, but if we start having constants in Assertion, maybe we don't need the Credible contract and instead have `ph` be part of the Assertion definition.

0xgregthedev: Yeah I would agree with this.

PR #15, opened 2025-02-17
Feat/state changes

fredo:
https://linear.app/phylaxsystems/issue/ENG-91/featcredible-std-add-state-changes-fns

A couple of notes: 

Introduced

Direct slot access: `getStateChanges<TYPE>(address contractAddress, bytes32 slot)`
Mapping access: `getStateChanges<TYPE>(address contractAddress, bytes32 slot, bytes32 key)`
Mapping access with offset (values might use multiple slots if structs): `getStateChanges<TYPE>(address contractAddress, bytes32 slot, bytes32 key, uint256 offset)`

### DIFF
--- a/src/Assertion.sol
+++ b/src/Assertion.sol
@@ -17,7 +17,7 @@ abstract contract Assertion is Credible {
         triggerRecorder.registerCallTrigger(fnSelector);
     }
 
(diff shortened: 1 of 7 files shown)


0xgregthedev (approved):
Looks good. 
Although I question if it's necessary for us to mask the types coming from our cheatcode.

PR #23, opened 2025-06-10
Feat/test cases

0xgregthedev:
Add test cases to be used as integration tests in op-talos and assex


fredo:
looks good overall, pls have a look at my chained PR which removes a bunch of unused imports and unnecesary inheritance. If this is good pls merge https://github.com/phylaxsystems/credible-std/pull/24 into your branch here first.

src/test-cases/precompiles/Load.sol
makemake-kbo, line 1: nit: this doesnt purely test load. not bad per-se but the reasoning behind these was to test cheatcodes in isolation (at least what i got from it)

makemake-kbo (approved):

src/test-cases/precompiles/Load.sol
0xgregthedev: I don't think that was the reasoning. It was to test the additional functionality that we have implemented. And testing the load cheatcode while fork switching is an edge case we should correctly handle

PR #29, opened 2025-08-01
gregory/eng 904 featcredible std add id to callinputs struct and make

0xgregthedev:
- **chore: add interfaces**
- **chore: remove mocks and update fork names throughout**
- **chore: consistent naming**


fredo (approved):
looks good

PR #59, opened 2026-04-30
feat: V2 syntax and assertions, protection suite examples

makemake-kbo:


fredo:
as discussed in the call 

- Address Aave bugs
- restructure royco example as circuit breaker example
- remove symbiotic

PR #66, opened 2026-05-15
feat: Add vault demo examples

makemake-kbo:


examples/vault_demo/src/VaultDemoAssertions.sol
```diff
@@ -0,0 +1,172 @@
...
+
+    function triggers() external view override {
+        watchCumulativeInflow(asset, THRESHOLD_BPS, INFLOW_WINDOW, this.assertCumulativeInflow.selector);
+        watchCumulativeOutflow(asset, THRESHOLD_BPS, OUTFLOW_WINDOW, this.assertCumulativeOutflow.selector);
+    }
+
+    function assertCumulativeInflow() external view {
+        PhEvm.InflowContext memory ctx = ph.inflowContext();
+        require(ph.getAssertionAdopter() == vault, "VaultDemo: wrong vault");
+        require(ctx.token == asset, "VaultDemo: wrong inflow token");
```
fredo: What are these two require statements doing? 
should vault always equal ph.assertionAdopter? then move the require into the constructor. 
what sets the ctx.token? If it's the asset you configure in triggers then it's always true. 

I see the danger here that misconfigured assertions might dos the vault becuase auf these require statements

examples/vault_demo/src/VaultDemoAssertions.sol
```diff
@@ -0,0 +1,172 @@
...
+
+        revert("VaultDemo: cumulative inflow breaker tripped");
+    }
+
+    function assertCumulativeOutflow() external view {
+        PhEvm.OutflowContext memory ctx = ph.outflowContext();
+        require(ph.getAssertionAdopter() == vault, "VaultDemo: wrong vault");
+        require(ctx.token == asset, "VaultDemo: wrong outflow token");
```
fredo: same here

fredo (changes requested):
looks mostly good, left one remark regarding require statements.

examples/vault_demo/src/VaultDemoAssertions.sol
makemake-kbo: https://github.com/phylaxsystems/credible-std/pull/66/commits/9810672b4514edbc8f9b65c82fd2b5514a365527

examples/vault_demo/src/VaultDemoAssertions.sol
makemake-kbo: https://github.com/phylaxsystems/credible-std/pull/66/commits/9810672b4514edbc8f9b65c82fd2b5514a365527

i dont think this is necessarily a dos but for sure a dead weight check

fredo (approved):

PR #68, opened 2026-05-21
feat(anomaly): anomalous tx detection interface

0xAlcibiades:
Adds interface for detecting and reacting to anomalous transactions. Notably omits test coverage for the moment until cheat codes land in `phorge`.

### DIFF
--- a/src/Assertion.sol
+++ b/src/Assertion.sol
@@ -133,6 +133,20 @@ abstract contract Assertion is ForkUtils, StateChanges {
         triggerRecorder.watchCumulativeInflow(token, thresholdBps, windowDuration, fnSelector);
     }
 
+    /// @notice Registers an anomaly-detection trigger. Fires whenever the
+    ///         executor's configured AnomalySubsystem produces a score for
+    ///         `target` in a transaction that touches it.
+    /// @dev The model owns the firing decision: the trigger fires whenever
+    ///      the subsystem returns a score at all. The assertion reads the
+    ///      score back via `ph.anomalyContext(target)` and decides whether
+    ///      to revert, run extra checks, or ignore.
+    /// @param target The address whose anomaly score this assertion observes.
+    /// @param fnSelector The assertion function to invoke when `target` is
+    ///        scored.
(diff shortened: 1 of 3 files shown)


src/PhEvm.sol
```diff
@@ -480,6 +480,26 @@ interface PhEvm {
     /// @return ctx The inflow context for the current trigger invocation.
     function inflowContext() external view returns (InflowContext memory ctx);
 
+    // ---------------------------------------------------------------
+    //  V2: Anomaly detection
```
makemake-kbo, line 484: Maybe we should have a new spec with anomaly detection? Wdyt cc: @odyslam @fredo

src/Assertion.sol
```diff
@@ -133,6 +133,20 @@ abstract contract Assertion is ForkUtils, StateChanges {
...
+    ///         executor's configured AnomalySubsystem produces a score for
+    ///         `target` in a transaction that touches it.
+    /// @dev The model owns the firing decision: the trigger fires whenever
+    ///      the subsystem returns a score at all. The assertion reads the
+    ///      score back via `ph.anomalyContext(target)` and decides whether
+    ///      to revert, run extra checks, or ignore.
+    /// @param target The address whose anomaly score this assertion observes.
+    /// @param fnSelector The assertion function to invoke when `target` is
+    ///        scored.
+    function watchAnomaly(address target, bytes4 fnSelector) internal view {
```
makemake-kbo, line 146: No way to specify what anomalies to specifically watch? I might have a wrong mental model of the system

makemake-kbo (approved):

src/Assertion.sol
0xAlcibiades: Nah, it's a single scalar score.

src/PhEvm.sol
fredo: I think it makes sense, since we are not gonna ship anomaly to Linea in the next upgrade

fredo (approved):
lgtm. Regarding the spec, I think we can follow up later once we implement the executor interface to a new spec.

PR #69, opened 2026-05-21
feat: Safe protection suite

makemake-kbo:


src/protection/safe/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+The pack exposes one assertion function per enforced policy: `assertSafeModulePolicy`, `assertSafeDelegateCallPolicy`, `assertSafeTargetSelectorPolicy`, `assertSafeBatchPolicy`, and `assertSafeApprovalPolicy`.
+
+Every watched execution is normalized into one or more actions. A normal owner or module transaction is one action. An approved `MultiSend` transaction is expanded into its packed inner actions, and every inner action is checked with the same rules.
+
+### What It Protects
+
+- Unknown contract targets with calldata are blocked.
+- Known targets must use an explicitly allowed selector unless the target is configured with the advanced `allowAnySelector` escape hatch.
+- Calldata shorter than four bytes is blocked unless the target explicitly allows empty or fallback-style calls.
```
fredo: how do you know this?

src/protection/safe/SafeConfigLockAssertion.sol
```diff
@@ -0,0 +1,97 @@
...
+///      is a sentinel meaning "modules must be disabled".
+contract SafeConfigLockAssertion is SafeConfigLockHelpers {
+    uint256 public immutable minThreshold;
+    uint256 public immutable minOwners;
+    address public immutable expectedGuard;
+    address public immutable expectedModuleGuard;
+    address public immutable expectedFallbackHandler;
+
+    bytes32[] public approvedOwnerSetHashes;
+    bytes32[] public approvedModuleSetHashes;
```
fredo, line 26: Can you pls double check if we also store storage with the asseriton after deployeing?

src/protection/safe/SafeConfigLockAssertion.sol
```diff
@@ -0,0 +1,97 @@
...
+
+        for (uint256 i; i < approvedModuleSetHashes_.length; ++i) {
+            approvedModuleSetHashes.push(approvedModuleSetHashes_[i]);
+        }
+
+        _registerReshiramSpec();
+    }
+
+    function triggers() external view override {
+        registerTxEndTrigger(this.assertSafeConfiguration.selector);
```
fredo: A slightly more optimized trigger here would be a state change trigger for the respective slots

src/protection/safe/SafeConfigLockAssertion.sol
makemake-kbo: we persist storage on deployment so we'll have the storage for every assertion run look the same as it would post deploy

src/protection/safe/README.md
makemake-kbo: yeah we check the calldata and allow based on how you configure the assertion but the readme makes it weird. will update

makemake-kbo: https://github.com/phylaxsystems/credible-std/pull/69/commits/f8b9039dee1fca34bb5e92b254f145ae48c9d675

src/protection/safe/SafeConfigLockAssertion.sol
makemake-kbo: https://github.com/phylaxsystems/credible-std/pull/69/commits/5c1e3df252d7ca4174ac6a4d1209aeec5fc46f8a

.github/workflows/solidity-test.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     uses: phylaxsystems/actions/.github/workflows/solidity-base.yaml@main
     with:
       gas-diff-tolerance: 25
-      forge-test-args: "--no-match-path 'test/backtesting/**'"
+      forge-test-args: "--no-match-path 'test/{backtesting,protection/safe}/**'"
```
fredo, line 13: why is safe excluded from tests?

makemake-kbo: it needs pcl test

src/protection/safe/SafeTxShapeHelpers.sol
```diff
@@ -0,0 +1,836 @@
...
+    error SafeTxShapeApprovalSpenderNotAllowed(address token, address spender, uint8 kind);
+    error SafeTxShapeApprovalUnlimitedBlocked(address token, address spender, uint8 kind);
+    error SafeTxShapeApprovalAmountAboveCap(
+        address token, address spender, uint8 kind, uint256 amount, uint256 maxAmount
+    );
+
+    TargetPolicy[] public targetPolicies;
+    SelectorPolicy[] public selectorPolicies;
+    BatchExecutorPolicy[] public batchExecutorPolicies;
+    ApprovalPolicy[] public approvalPolicies;
```
fredo, line 136: `approvalPolicies` is stored as an array and each enforcement path currently does a linear scan over all policies. That is acceptable for a small pack, but if a Safe has many tokens/spenders/operators this becomes O(n) work on every approval-like action.

Two reasonable lookup layouts:

- `mapping(bytes32 => ApprovalPolicy)` where `key = keccak256(abi.encode(token, spender, kind))`. This keeps the index compact and generic, but it is less self-documenting.
- `mapping(address => mapping(address => mapping(uint8 => ApprovalPolicy)))` plus an `exists` flag. This is more explicit and avoids composite-key hashing; I would lean this way for `(token, spender, kind)`.

You can still keep the array for enumeration/public inspection and populate the mapping in the constructor for constant-time enforcement.

src/protection/safe/SafeTxShapeHelpers.sol
```diff
@@ -0,0 +1,836 @@
...
+            if (action.dataLength != 68) revert SafeTxShapeApprovalMalformed(action.target, selector);
+            if (!_tokenHasApprovalKind(action.target, APPROVAL_KIND_ERC20_INCREASE_ALLOWANCE)) {
+                revert SafeTxShapeApprovalTokenUnconfigured(action.target, selector);
+            }
+
+            address spender = _readAbiAddress(action.data, action.dataOffset + 4);
+            uint256 addedValue = _readUint256(action.data, action.dataOffset + 36);
+            if (addedValue == 0) return;
+
+            _validateNumericApproval(action.target, spender, APPROVAL_KIND_ERC20_INCREASE_ALLOWANCE, addedValue);
```
fredo: Potential correctness issue: `increaseAllowance` is not the same shape as `approve`. For the common OpenZeppelin-compatible behavior, `increaseAllowance(spender, addedValue)` adds to the current allowance; it does not set the final allowance to `addedValue`.

With `maxAmount = 100`, if current allowance is already `90`, `increaseAllowance(spender, 100)` passes this check because `100 <= maxAmount`, but final allowance becomes `190`. Same in a MultiSend with two `increaseAllowance(spender, 100)` calls: each inner action passes individually, but final allowance can become `200`.

If `maxAmount` is meant to be a final allowance cap, this path should read/check post-state `allowance(safe, spender) <= maxAmount`, or the README should call this a per-call delta cap. The safer MVP may be to disallow `increaseAllowance` unless final allowance is verified.

src/protection/safe/SafeTxShapeHelpers.sol
makemake-kbo: https://github.com/phylaxsystems/credible-std/pull/69/commits/7baa2889bcabbb5e7810c201e07a321e04163dd8

src/protection/safe/SafeTxShapeHelpers.sol
makemake-kbo: https://github.com/phylaxsystems/credible-std/pull/69/commits/7baa2889bcabbb5e7810c201e07a321e04163dd8

Bound the cap to the post-state `allowance(safe, spender)` via `staticcallAt` on the PostCall fork so stacked `increaseAllowance` calls inside a MultiSend can't land above `maxAmount`. README updated to spell out approve vs. increaseAllowance cap semantics.

.github/workflows/solidity-test.yml
mateo-mro, line 13: tagging on top of fredo's earlier thread — the exclusion is still here and the reply was "it needs pcl test". whatever "needs pcl test" resolves to, merging in this state ships ~1k lines of tests under `test/protection/safe/**` that nobody is running. if the tests can't run in this workflow, either (a) move them to whichever workflow can run them now and add that lane to required checks, or (b) drop them from the PR and bring them back with the pcl support. otherwise the next person who touches these contracts has no signal that the tests are dark.

src/protection/safe/SafeConfigLockHelpers.sol
```diff
@@ -0,0 +1,119 @@
...
+        address next;
+        (modules, next) = abi.decode(
+            _viewAt(
+                safe,
+                abi.encodeCall(ISafeConfigLockTarget.getModulesPaginated, (SENTINEL_MODULES, MODULE_PAGE_SIZE)),
+                fork
+            ),
+            (address[], address)
+        );
+        require(next == SENTINEL_MODULES, "SafeConfigLock: too many modules");
```
mateo-mro: `require(next == SENTINEL_MODULES, …)` fails the assertion entirely if a Safe has more than `MODULE_PAGE_SIZE = 256` modules. that means every triggering tx on such a Safe gets blocked — not because policy is violated, but because the helper can't read the full list. multisigs that use modules heavily (especially with auto-enabled session keys or recovery modules) can realistically blow past 256.

either paginate (loop `getModulesPaginated(next, MODULE_PAGE_SIZE)` until `next == SENTINEL_MODULES`) or document the 256-module cap as a hard constraint in the README. silent fail-closed is the worst of both worlds.

src/protection/safe/SafeConfigLockHelpers.sol
```diff
@@ -0,0 +1,119 @@
...
+        0x6c9a6c4a39284e37ed1cf53d337577d14212a4870fb976a4366c693b939918d5;
+    bytes32 internal constant GUARD_STORAGE_SLOT = 0x4a204f620c8c5ccdca3fd54d003badd85ba500436a431f0cbda4f558c93c34c8;
+    bytes32 internal constant MODULE_GUARD_STORAGE_SLOT =
+        0xb104e0b93118902c651344349b610029d694cfdec91c589c91ebafbcd0289947;
+
+    /// @notice Computes the deterministic hash used by owner and module allow lists.
+    /// @dev Sorts the provided addresses in memory before hashing, so Safe linked-list order
+    ///      does not affect the resulting set hash.
+    function hashAddressSet(address[] memory accounts) public pure returns (bytes32) {
+        _sortAddresses(accounts);
```
mateo-mro: `hashAddressSet` is `public pure` but sorts `accounts` in place via `_sortAddresses`. memory arrays in solidity are passed by reference, so any caller who hands in their own `owners` / `modules` array (e.g. a deploy script reusing the array after computing the hash) will find it mutated. either copy into a local before sorting, or rename / document the side effect (something like `"hashAddressSet sorts the input array as a side-effect"` in the docstring). bonus: the in-place insertion sort gives no benefit when callers are obligated to discard the input anyway, so a copy is essentially free.

src/protection/safe/SafeConfigLockHelpers.sol
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    function _viewFailureMessage() internal pure override returns (string memory) {
+        return "SafeConfigLock: safe view failed";
+    }
+
+    function _registerReshiramSpec() internal {
+        (bool ok,) = SPEC_RECORDER.call(
+            abi.encodeWithSelector(bytes4(keccak256("registerAssertionSpec(uint8)")), AssertionSpec.Reshiram)
+        );
```
mateo-mro: `(bool ok,) = SPEC_RECORDER.call(...)` drops the returned bytes, so when registration fails the `"spec registration failed"` require gives ops nothing to grep on. also: building the selector via `bytes4(keccak256("registerAssertionSpec(uint8)"))` instead of calling the typed interface means a future signature change won't surface at compile time, only at deploy. would feel safer to (a) cast `SPEC_RECORDER` to the registrar interface and call it directly, and (b) forward the returned data on failure (`require(ok, string(returndata))` or a custom error that carries the bytes).

src/protection/safe/SafeTxShapeHelpers.sol
```diff
@@ -0,0 +1,865 @@
...
+        }
+    }
+
+    function _validateIncreaseAllowanceFinalState(address safe, address token, address spender) internal view {
+        if (!_approvalPolicyExists[token][spender][APPROVAL_KIND_ERC20_INCREASE_ALLOWANCE]) {
+            revert SafeTxShapeApprovalSpenderNotAllowed(token, spender, APPROVAL_KIND_ERC20_INCREASE_ALLOWANCE);
+        }
+
+        PhEvm.TriggerContext memory triggerCtx = ph.context();
+        PhEvm.ForkId memory postFork = PhEvm.ForkId({forkType: 3, callIndex: triggerCtx.callEnd});
```
mateo-mro: this constructs `PhEvm.ForkId({forkType: 3, callIndex: triggerCtx.callEnd})` by hand instead of calling `_postCall(triggerCtx.callEnd)`. `_postCall` exists for exactly this purpose (`src/Assertion.sol:172`). hard-coding `forkType: 3` is brittle — if the `ForkType` enum is ever renumbered upstream, this line silently reads from the wrong fork and the increaseAllowance cap stops doing what the name says. please route through `_postCall`.

src/protection/safe/SafeTxShapeHelpers.sol
```diff
@@ -0,0 +1,865 @@
...
+    bytes4 public constant EXEC_TRANSACTION_FROM_MODULE_RETURN_DATA_SELECTOR =
+        bytes4(keccak256("execTransactionFromModuleReturnData(address,uint256,bytes,uint8)"));
+
+    bytes4 public constant MULTISEND_SELECTOR = bytes4(keccak256("multiSend(bytes)"));
+    bytes4 public constant APPROVE_SELECTOR = bytes4(keccak256("approve(address,uint256)"));
+    bytes4 public constant INCREASE_ALLOWANCE_SELECTOR = bytes4(keccak256("increaseAllowance(address,uint256)"));
+    bytes4 public constant SET_APPROVAL_FOR_ALL_SELECTOR = bytes4(keccak256("setApprovalForAll(address,bool)"));
+
+    uint256 internal constant MULTISEND_HEADER_LENGTH = 85;
+    uint64 internal constant ALLOWANCE_READ_GAS = 500_000;
```
mateo-mro, line 34: `ALLOWANCE_READ_GAS = 500_000` hardcoded. for a vanilla ERC-20 it's overkill; for a proxied / hooked ERC-20 whose `allowance(address,address)` legitimately needs more (some L2 wrappers, account-abstraction tokens, fee-hooked tokens), the staticcall hits the cap and returns `ok == false`. that then falls into `revert SafeTxShapeApprovalMalformed(token, INCREASE_ALLOWANCE_SELECTOR)` on line 511 — which blocks the user's tx, with an error message that's not actually accurate (the calldata is fine; the read failed).

consider either making this configurable per-policy, or at minimum splitting the error into `SafeTxShapeAllowanceReadFailed(token)` so ops can distinguish a real malformed approval from a gas-starved read.

src/protection/safe/SafeTxShapeHelpers.sol
```diff
@@ -0,0 +1,865 @@
...
+        internal
+        view
+        returns (bool)
+    {
+        if (dataLength < 4) return false;
+        (bool found,) = _batchPolicyIndex(target, _selectorAt(data, dataOffset));
+        return found;
+    }
+
+    function _targetPolicyIndex(address target) internal view returns (bool found, uint256 index) {
```
mateo-mro, line 681: fredo flagged the O(n) `approvalPolicies` scan earlier and you moved that to a mapping — but the same pattern is still here in `_targetPolicyIndex`, `_selectorPolicyIndex` (line 655), `_batchPolicyIndex` (662), and `_validateModuleCaller` (541). every triggered Safe call pays a linear scan for each of these.

for a typical Safe with a handful of targets + selectors + one MultiSend executor + one or two modules, it's nothing. for a Safe configured to talk to a large protocol surface (e.g. a treasury Safe wired to 20+ contracts × multiple selectors) the per-tx assertion gas grows quadratically with the cross-product. easy follow-up: keep the storage arrays for enumeration but back them with `mapping(address => uint256) targetPolicyIndexPlusOne` (and equivalents), same shape as `_approvalPolicyByKey`.

src/protection/safe/SafeTxShapeHelpers.sol
```diff
@@ -0,0 +1,865 @@
...
+                    callEnd: context.callEnd
+                });
+            }
+        }
+
+        revert SafeTxShapeTriggeredCallNotFound(context.selector, context.callStart);
+    }
+
+    function _decodeOwnerTx(bytes memory input) internal pure returns (OwnerTx memory ownerTx) {
+        if (input.length < 324 || _selector(input) != EXEC_TRANSACTION_SELECTOR) {
```
mateo-mro: `input.length < 324` is below the minimum valid `execTransaction` encoding. minimum is `4 (selector) + 10*32 (head) + 32 (data length word) + 32 (signatures length word) = 388` bytes (with empty `data` and empty `signatures`). inputs in the 324–387 range pass this check but will revert later inside `_readDynamicBytes` with the same `SafeTxShapeBatchPayloadMalformed` — just one layer deeper.

functionally fail-closed, but if you bump the bound to 388 the early-exit error is more precise. same likely applies to the `< 132` check on `_decodeModuleTx` two lines below — the floor for `execTransactionFromModule[ReturnData]` with empty data is `4 + 4*32 + 32 = 164`, not 132.

src/protection/safe/SafeTxShapeHelpers.sol
```diff
@@ -0,0 +1,865 @@
...
+        }
+
+        PhEvm.TriggerContext memory triggerCtx = ph.context();
+        PhEvm.ForkId memory postFork = PhEvm.ForkId({forkType: 3, callIndex: triggerCtx.callEnd});
+
+        PhEvm.StaticCallResult memory result = ph.staticcallAt(
+            token, abi.encodeWithSignature("allowance(address,address)", safe, spender), ALLOWANCE_READ_GAS, postFork
+        );
+        if (!result.ok || result.data.length < 32) {
+            revert SafeTxShapeApprovalMalformed(token, INCREASE_ALLOWANCE_SELECTOR);
```
mateo-mro: the failure mode here is "`allowance(safe, spender)` staticcall returned `!ok` or returned <32 bytes" — i.e. the token's view function failed, not the user's approval calldata. reverting with `SafeTxShapeApprovalMalformed` mis-labels that for any operator staring at the failed-assertion event. consider a distinct error like `SafeTxShapeAllowanceReadFailed(token, spender)` so incident response can tell "user sent garbage" apart from "token contract misbehaved / ran out of gas".

fredo (dismissed):
lgtm

fredo (approved):

PR #70, opened 2026-05-25
feat: consolidate per-protocol assertion examples under examples/

makemake-kbo:
Brings assertion examples from the per-protocol branches (aave, aerodrome, cap, curve, denaria, euler, nado, royco, safe, spark, symbiotic, tydro, uniswap, veda, zeroex) into examples/<protocol>/src/ on a single branch.

Each protocol is compiled standalone via a new [profile.<protocol>] in foundry.toml that shares the root src/ (credible-std) and lib/ via the credible-std/ remapping — no nested foundry projects, no duplicated submodules. Imports in extracted files are rewritten to the credible-std/ remapping form.

Root src/ is extended with the missing protection bases that examples depend on (access_control/*, perpetual/PerpetualBaseAssertion + IPerpetualProtectionSuite). AaveV3LikeOperationSafety is upgraded to the full version that exposes AaveV3LikeProtectionSuite. PhEvm gains the marketPrice precompile signature and IAaveV3LikeOracle gains BASE_CURRENCY / BASE_CURRENCY_UNIT.

All 20 foundry profiles build clean.


fredo (commented):
Constructor-time protocol reads need to be removed before these examples are deployable as Credible assertions.

examples/curve/src/CurveLlammaProtocol.sol
```diff
@@ -0,0 +1,128 @@
...
+    constructor(
+        address amm_,
+        uint256 borrowedPrecision_,
+        uint256 collateralPrecision_,
+        uint256 maxBandsToScan_,
+        uint256 dustTolerance_,
+        uint256 priceTolerance_
+    ) {
+        amm = amm_;
+        borrowedToken = ICurveLlammaAMM(amm_).coins(0);
```
fredo: This constructor reads live AMM state (`coins`) and then token metadata (`decimals`) below. That has the same deployability problem documented in the ERC4626/Uniswap helpers: assertion construction runs in an isolated deploy context, so adopter/token reads can revert with `EXTCODESIZE = 0`. Please pass the borrowed token, collateral token, and precisions/decimals explicitly, or move reads to fork-time helpers using `ph.staticcallAt`.

examples/curve/src/CurveUsdProtocol.sol
```diff
@@ -0,0 +1,192 @@
...
+        bytes4 selector;
+        address caller;
+        bytes input;
+        uint256 callStart;
+        uint256 callEnd;
+    }
+
+    constructor(address controller_, uint256 maxLoansToScan_, uint256 debtTolerance_) {
+        controller = controller_;
+        amm = ICurveUsdController(controller_).amm();
```
fredo: Same constructor-read issue here: `controller_.amm()` reaches into the protocol while the assertion is being constructed. Please pass the AMM address explicitly, or derive it later via a fork-aware read.

examples/curve/src/LlamaLendProtocol.sol
```diff
@@ -0,0 +1,116 @@
...
+    bytes4 internal constant BORROW_MORE_CALLBACK = bytes4(keccak256("borrow_more(uint256,uint256,address,address)"));
+    bytes4 internal constant BORROW_MORE_CALLBACK_DATA =
+        bytes4(keccak256("borrow_more(uint256,uint256,address,address,bytes)"));
+}
+
+abstract contract LlamaLendVaultProtocolHelpers is ERC4626AssetFlowAssertion {
+    address internal immutable controller;
+
+    constructor(address vault_) {
+        controller = ILlamaLendVault(vault_).controller();
```
fredo: Same issue: this constructor reads `vault_.controller()` from the adopter/protocol. Constructors should only consume deployment args for this kind of dependency; pass the controller explicitly or read it later through the fork-aware helpers.

examples/curve/src/LlamaLendProtocol.sol
```diff
@@ -0,0 +1,116 @@
...
+}
+
+abstract contract LlamaLendControllerProtocolHelpers is Assertion {
+    address internal immutable controller;
+    address internal immutable borrowedToken;
+    uint256 internal immutable availableBalanceTolerance;
+
+    constructor(address controller_, uint256 availableBalanceTolerance_) {
+        controller = controller_;
+        borrowedToken = ILlamaLendController(controller_).borrowed_token();
```
fredo: Same constructor-read issue: `controller_.borrowed_token()` depends on live protocol code during assertion deployment. Please pass the borrowed token explicitly or defer this through `ph.staticcallAt` at assertion execution time.

examples/royco/src/RoycoHelpers.sol
```diff
@@ -0,0 +1,446 @@
...
+    address internal immutable kernel;
+    address internal immutable accountant;
+    address internal immutable seniorTranche;
+    address internal immutable juniorTranche;
+    address internal immutable stAsset;
+    address internal immutable jtAsset;
+
+    constructor(address kernel_) {
+        kernel = kernel_;
+        accountant = IRoycoKernel(kernel_).ACCOUNTANT();
```
fredo: This constructor reads multiple values from the kernel (`ACCOUNTANT`, tranches, and assets). That can fail in the assertion-deploy runtime for the same reason as the other adopter reads. Please pass these dependencies explicitly or make them fork-time reads.

examples/royco/src/RoycoHelpers.sol
```diff
@@ -0,0 +1,446 @@
...
+/// @author Phylax Systems
+/// @notice Consolidated Royco tranche-side read and decode helpers for assertions.
+abstract contract RoycoVaultTrancheHelpers is RoycoHelpers {
+    address internal immutable tranche;
+    address internal immutable kernel;
+    RoycoTrancheType internal immutable trancheType;
+
+    constructor(address tranche_) {
+        tranche = tranche_;
+        kernel = IRoycoVaultTranche(tranche_).KERNEL();
```
fredo: This constructor also reads protocol state from the tranche (`KERNEL` and `TRANCHE_TYPE`). Please avoid live protocol calls during construction; pass these values explicitly or defer them to fork-aware reads.

examples/spark/src/SparkLendOraclePriceGuardAssertion.sol
```diff
@@ -0,0 +1,300 @@
...
+    /// @param pool_ SparkLend pool whose risk-increasing selectors are monitored.
+    /// @param oracle_ AaveOracle used by the pool to report asset prices.
+    /// @param watchEntries_ Per-asset market-pair and tolerance configuration.
+    constructor(address pool_, address oracle_, WatchEntry[] memory watchEntries_) {
+        require(pool_ != address(0), "SparkLendOracle: zero pool");
+        require(oracle_ != address(0), "SparkLendOracle: zero oracle");
+
+        pool = pool_;
+        oracle = oracle_;
+        baseCurrency = IAaveV3LikeOracle(oracle_).BASE_CURRENCY();
```
fredo: This reads oracle configuration during construction (`BASE_CURRENCY`, and `BASE_CURRENCY_UNIT` on the next line). To keep the assertion deployable in the Credible runtime, these should be constructor arguments or fork-time reads rather than live oracle calls in the constructor.

examples/symbiotic/src/SymbioticVaultBaseAssertion.sol
```diff
@@ -0,0 +1,24 @@
...
+    /// @notice The Symbiotic vault being monitored.
+    address internal immutable vault;
+
+    /// @notice The ERC-20 collateral backing the vault.
+    address internal immutable asset;
+
+    constructor(address vault_) {
+        require(vault_ != address(0), "SymbioticVaultBase: vault is zero");
+        vault = vault_;
+        asset = ISymbioticVaultLike(vault_).collateral();
```
fredo: This constructor reads `vault_.collateral()` from the protected vault. Please pass the collateral asset explicitly, or read it later through the fork-aware helper path, so assertion deployment does not depend on live adopter code being available in the constructor context.

src/PhEvm.sol
```diff
@@ -528,4 +528,10 @@ interface PhEvm {
         external
         pure
         returns (bool);
+
+    /// @notice Returns the market price of `asset` quoted in `denomAsset`, normalized to 18 decimals.
+    /// @param asset The asset whose price is requested.
+    /// @param denomAsset The quote asset, or address(0) for USD.
+    /// @return price The 18-decimal market price of `asset` in `denomAsset`.
+    function marketPrice(address asset, address denomAsset) external view returns (uint256 price);
```
mateo-mro: this declares the `marketPrice` precompile but i can't find a corresponding implementation anywhere in `credible-sdk` (grepped `crates/assertion-executor` for `marketPrice` / `market_price` — zero hits, whereas `conserveBalance`, `forbidChangeForSlots`, `assetsMatchSharePrice` all resolve to files under `crates/assertion-executor/src/inspectors/precompiles/reshiram/`).

that means anything in this PR that calls `ph.marketPrice(...)` will revert at assertion-execution time with no clear signal that the precompile just doesn't exist yet — see `SparkLendOraclePriceGuardAssertion._checkEntry` on line 182 of `examples/spark/src/SparkLendOraclePriceGuardAssertion.sol`.

could you either (a) hold the signature back until the sidecar side is merged, (b) gate the spark example behind a feature flag / readme caveat saying "requires sidecar release X.Y", or (c) point at the tracking issue if there is one? right now an adopter following the readme will deploy a bundle whose risk-increasing path always reverts.

examples/spark/src/SparkLendOraclePriceGuardAssertion.sol
```diff
@@ -0,0 +1,315 @@
...
+            (bool initialized, AaveV3LikeTypes.ReserveData memory reserveData) = _tryGetReserveData(entry.asset, fork);
+            if (initialized && _isUsingReserve(userConfig.data, reserveData.id)) {
+                _checkEntry(entry, fork);
+            }
+        }
+    }
+
+    function _checkEntry(WatchEntry memory entry, PhEvm.ForkId memory fork) internal view {
+        uint256 reportedInDenom = _reportedPriceInDenom(entry.asset, entry.denomAsset, fork);
+        uint256 market = ph.marketPrice(entry.asset, entry.denomAsset);
```
mateo-mro: consumer side of the same concern: this is the only call site of `ph.marketPrice` in the PR, and the precompile isn't implemented in credible-sdk yet. on a real deployment this assertion's `assertOraclePricesTrackMarket` will revert on every triggered call.

at minimum, a doc comment at the top of the contract noting it depends on an unreleased precompile. ideally hold the example back until the runtime side lands or wire it to an externally-deployed market oracle (the same pattern `examples/micro-patterns/src/OracleReduceOnlyGate.a.sol` uses with `MARKET_REFERENCE` / `_readUintAt`).

src/protection/lending/examples/AaveV3LikeOperationSafety.sol
```diff
@@ -1,18 +1,471 @@
...
+        });
+    }
+
+    /// @notice Builds the liquidation collateral-consumption check from actual collateral transfers.
+    function _getLiquidationCollateralCheck(
+        OperationContext calldata operation,
+        PhEvm.ForkId calldata beforeFork,
+        PhEvm.ForkId calldata afterFork
+    ) internal view returns (ConsumptionCheck memory check) {
+        // bug: different decimals break return asset accounting
```
mateo-mro: `// bug: different decimals break return asset accounting` — this is shipping with the bug acknowledged in-code. that's a non-trivial spot: `_getLiquidationCollateralCheck` is what protects users from liquidation collateral being seized in excess of what `lossCreated` justifies. if the math is wrong under different-decimal asset pairs (which is the common case — most pools have collateral and debt with different decimals), the resulting `LIQUIDATION_COLLATERAL_CHECK` will either misfire or silently pass when it shouldn't.

either describe the exact conditions that trigger the bug here so an integrator can decide whether their pool is affected, or short-circuit the check (return `checks` length 0) with a clearer comment until it's actually fixed. shipping silent wrong-math under a misleading check name is worse than not running the check.

src/protection/lending/examples/AaveV3LikeOperationSafety.sol
```diff
@@ -1,18 +1,471 @@
...
+
+            if (!include) {
+                continue;
+            }
+
+            balances[count++] = balance;
+        }
+
+        assembly {
+            mstore(balances, count)
```
mateo-mro: trimming `balances.length` via `mstore` works but it leaves the slots beyond `count` still allocated in memory — if anyone ever changes this function to return `balances` over an external boundary, or if the `AccountBalance` struct layout changes, that trailing data is observable. would feel safer either (a) doing a second-pass copy into `new AccountBalance[](count)`, or (b) tracking down the `LendingProtectionSuiteBase` contract that consumes this array and confirming it never reads past `length` (which is the implicit assumption you're relying on here). either way please add a brief comment locking in the invariant.

src/protection/access_control/BalanceConservationAssertion.sol
```diff
@@ -0,0 +1,69 @@
...
+    /// @dev Uses registerTxEndTrigger so the check fires once after the transaction completes.
+    ///      Call this inside your `triggers()`.
+    function _registerBalanceConservationTriggers() internal view {
+        registerTxEndTrigger(this.assertBalanceConservation.selector);
+    }
+
+    /// @notice Verifies that all protected account balances are unchanged across the transaction.
+    /// @dev Checks each (token, account) pair using the `conserveBalance` precompile at
+    ///      PreTx vs PostTx. Reverts on the first pair whose balance changed.
+    function assertBalanceConservation() external {
```
mateo-mro: `external` without `view`. the body only reads via precompiles and never writes — i think this can (and should) be `external view`. same for `SharePriceAssertion.assertSharePrice` and `SlotProtectionAssertion.assertSlotProtection`. without the marker the compiler can't enforce the no-state-write guarantee, and tooling that inspects assertion signatures (slither, etc.) won't classify these as pure checks.

src/protection/access_control/BalanceConservationAssertion.sol
```diff
@@ -0,0 +1,69 @@
...
+        ConservedBalance[] memory balances = _conservedBalances();
+        PhEvm.ForkId memory pre = _preTx();
+        PhEvm.ForkId memory post = _postTx();
+
+        for (uint256 i = 0; i < balances.length; i++) {
```
mateo-mro, line 67: this is N precompile calls in a loop with no batch alternative — each `conserveBalance` does fork-time reads at both pre and post. for a protocol protecting many treasury/escrow pairs (say a multi-token vault with treasury + insurance pool + fee account), this scales linearly into the assertion gas budget. SlotProtection batches into a single `forbidChangeForSlots` call below for the same reason.

would either (a) a doc note saying "keep under ~N pairs for a normal gas budget", or (b) a follow-up to add a batch `conserveBalances(pre, post, (token,account)[])` precompile, save adopters from running into OOG reverts only at scale? right now nothing about the API surface hints at the per-pair cost.

src/protection/access_control/BalanceConservationAssertion.sol
```diff
@@ -0,0 +1,69 @@
...
+            require(
+                ph.conserveBalance(pre, post, balances[i].token, balances[i].account),
+                "AccessControl: protected balance changed"
```
mateo-mro: `"AccessControl: protected balance changed"` doesn't tell ops which (token, account) pair tripped. when you're staring at an assertion-failed event in a prod incident timeline that's the first thing you want. consider a custom error `BalanceChanged(address token, address account)` instead — same for the slot-protection / share-price messages.

src/protection/access_control/SharePriceAssertion.sol
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /// @notice Verifies that all protected vault share prices are stable across the transaction.
+    /// @dev Uses `ph.assetsMatchSharePrice()` for each vault, which reads totalAssets() and
+    ///      totalSupply() at every fork point and checks for deviation beyond the tolerance.
+    ///      Reverts on the first vault whose share price moved beyond tolerance.
+    function assertSharePrice() external {
+        ProtectedVault[] memory vaults = _protectedVaults();
+
+        for (uint256 i = 0; i < vaults.length; i++) {
```
mateo-mro, line 64: same per-vault-in-a-loop concern as `BalanceConservationAssertion` — `assetsMatchSharePrice` already does an all-forks check internally, so multiplying that by N vaults can grow the per-tx assertion cost quickly. fine for the 1-2 vault case but worth a doc note for anyone protecting a whole vault family.

src/protection/access_control/AccessControlBaseAssertion.sol
```diff
@@ -0,0 +1,35 @@
...
+/// }
+/// ```
+abstract contract AccessControlBaseAssertion is Assertion {
+    /// @notice The contract whose access control is being protected (assertion adopter).
+    address internal immutable target;
+
+    constructor(address _target) {
+        target = _target;
+    }
+}
```
mateo-mro, line 37: the composition pattern in your example block (inherit `SlotProtectionAssertion` + `BalanceConservationAssertion`, call `_registerSlotProtectionTriggers()` and `_registerBalanceConservationTriggers()` in `triggers()`) is silent-fail by design: if a future maintainer adds a third mixin and forgets the register call, that whole check class never runs and nothing surfaces it.

could you add a one-liner in the docstring above making the failure mode explicit ("forgetting `_register*Triggers()` for an inherited mixin disables that check silently"), or even better, expose a `_registerAllProtections()` helper the mixins push into via a hook so the implementer doesn't have to know each mixin's name?

src/protection/perpetual/PerpetualBaseAssertion.sol
```diff
@@ -0,0 +1,487 @@
...
+
+    /// @notice Internal implementation shared by the public perpetual assertion entrypoints.
+    function _assertOperationSafety() internal view {
+        IPerpetualProtectionSuite suite = _suite();
+        IPerpetualProtectionSuite.TriggeredCall memory triggered = _resolveTriggeredCall();
+        IPerpetualProtectionSuite.OperationContext memory operation = suite.decodeOperation(triggered);
+        PhEvm.ForkId memory beforeFork = _preCall(triggered.callStart);
+        PhEvm.ForkId memory afterFork = _postCall(triggered.callEnd);
+
+        _assertExecutionPriceChecks(suite, triggered, operation, beforeFork, afterFork);
```
mateo-mro: every triggered call fans out into 6 separate suite methods (execution-price, liquidity-coverage, funding-delta, liquidation, oracle-anchor, accounting-conservation) plus a post-mutation snapshot. for a suite that overrides only `getPostMutationSnapshot` (the common case from `PerpetualProtectionSuiteBase` defaults), the other 5 still pay the cross-contract call + abi-decode + empty-array allocation cost on every trigger.

would it be worth letting the suite advertise which check kinds it actually populates — e.g. a bitset returned alongside `getMonitoredSelectors`, or a separate `enabledCheckKinds()` view — so the base can skip the empty fanout? not blocking, but for perp protocols with high tx frequency this overhead adds up.

foundry.toml
```diff
@@ -42,6 +42,115 @@ remappings = [
   "credible-std/=src/"
 ]
 
+# Per-protocol example profiles.
+# Each profile compiles examples/<protocol>/src/ standalone, sharing root
+# credible-std (src/) and root lib/ via the credible-std remapping.
+
+[profile.aave]
```
mateo-mro, line 53: 16 new per-protocol profiles, and the PR description says "all 20 foundry profiles build clean" — is there a CI job actually iterating `FOUNDRY_PROFILE=<protocol> forge build` for each? i checked the workflow files in this PR and didn't spot one. without that, future PRs that touch root `src/` can quietly break example builds and nobody notices until an adopter tries to deploy. a single matrix step would catch it: `strategy.matrix.profile: [aave, aerodrome, cap, …]` running `forge build` per profile.

fredo (approved):
LGTM

PR #72, opened 2026-06-04
feat: optimize assertion triggers (onFnCall → onTxEnd where tx-wide)

makemake-kbo:
## Summary

Reduces how often Credible assertions fire **only where doing so preserves the exact invariant**. After review, the only safe reductions left are: keeping per-call where per-call boundaries matter, narrowing the slot-protection trigger to the protected slots, and running the protection suite under `pcl` in CI.

### What landed

- **Slot protection (`SlotProtectionAssertion`)**: register one `registerStorageChangeTrigger(assertSlotProtection, slot)` per protected slot instead of a generic tx-end trigger. The check now runs only when a watched slot is actually written, narrowing execution to the exact risk surface. Coverage is preserved because `forbidChangeForSlots` inspects the same transaction journal a storage-change trigger fires on, so same-value writes are still caught.
- **Lending decode fix**: `ph.getAllCallInputs(...).input` returns calldata args **without** the 4-byte selector, but the arg decoders expect selector-prefixed calldata (`input[4:]`). The shared lending base fed the no-selector input straight into `decodeOperation`, decoding the wrong account at runtime. The selector is now prepended in the enumerate path. (Unit tests missed this because they build input via `abi.encodeCall`, which includes the selector.) `LendingSolvencyPerCall.t.sol` covers the per-call path: honest-pass, breaking-op-trips, pre-insolvent-skipped, and transient-insolvency-repaired-still-trips.
- **CI**: the whole `test/protection/**` suite runs under `pcl test` (the `cl.assertion` cheatcode is Phorge-only), not base `forge test`.

### Reverted after review

An earlier revision moved several checks from `registerFnCallTrigger` (onFnCall) to `registerTxEndTrigger`. Review ([thread](https://github.com/phylaxsystems/credible-std/pull/72#discussion_r3358642125) and [follow-up](https://github.com/phylaxsystems/credible-std/pull/72#issuecomment-4632979840)) established that none of those conversions were strictly coverage-preserving — a per-account or per-mutation state that is bad mid-tx and repaired before tx end would no longer be flagged — so **all of them are reverted to per-call**:

- **Generic lending** post-operation solvency (`LendingBaseAssertion`).
- **AaveV4 Spoke**: `assertAccountDataMatchesIndependentState`, `assertLiquidationImprovesBorrowerRisk`.
- **crvUSD**: `assertPostRiskIncreasingOperationHealth`, `assertPostLiquidationHealthRule`.
- **LlamaLend**: `assertDebtIncreaseWithinBorrowCap`.
- **AaveV4 Hub**: `assertHubAssetAccounting` (restores per-call `assetId` filter + per-mutation monotonic checks).
- **Aerodrome**: `assertReservesBackedByBalances`.
- **UniV3**: `assertOracleStateConsistent`, `assertProtocolFeesCoveredByCustody`.

The tx-end enumerate helpers and the tx-end-specific example tests written for those conversions were removed.

## Testing

All via `pcl test` (the `cl.assertion` cheatcode is Phorge-only). Green across aave / aerodrome / curve / uniswap profiles and the root `test/protection/**` suite under `FOUNDRY_PROFILE=ci`, plus base `forge test`.


src/protection/lending/LendingBaseAssertion.sol
```diff
@@ -98,44 +109,138 @@ abstract contract LendingBaseAssertion is Assertion {
...
-        bytes4[] memory selectors = _suite().getMonitoredSelectors();
-        for (uint256 i; i < selectors.length; ++i) {
-            registerFnCallTrigger(this.assertOperationSafety.selector, selectors[i]);
+        ILendingProtectionSuite suite = _suite();
+
+        bytes4[] memory consumptionSelectors = suite.getConsumptionSelectors();
+        for (uint256 i; i < consumptionSelectors.length; ++i) {
+            registerFnCallTrigger(this.assertOperationSafety.selector, consumptionSelectors[i]);
         }
+
```
fredo: I’m not sure lending solvency should move to a tx-end-only check. The old per-call path caught the exact call that first made a previously solvent account insolvent; the new `assertAccountSolvency()` only compares PreTx vs PostTx for deduped accounts.

That means a transaction can borrow/remove collateral into insolvency, then later repay/add collateral before tx end, and the assertion passes. If the intended invariant is “risk-increasing lending operations must not create an insolvent intermediate account state,” this should stay per-call, or we need both checks: per-call solvency for risk-increasing operations plus tx-end dedupe as an optimization for aggregate final-state coverage.

makemake-kbo: yeah ive thought about this and imo a value extracting call *could* be defined as explicitly post state extracting, and with the goal of reducing triggers i decided to move it to post call. i will move it back

makemake-kbo: Good catch — agreed, reverted to per-call in 1862557.

`assertOperationSafety` again runs post-operation solvency per call on every monitored selector (alongside the bounded-consumption checks), so the exact call that first makes a previously solvent account insolvent is the one that reverts. The tx-end trigger, `assertAccountSolvency`/`_assertAccountSolvency`, the `LendingAccountSolvencyViolated` error, and the `getConsumptionSelectors` narrowing (which only made sense once solvency lived at tx end) are all gone. The `getAllCallInputs` selector-prepend fix in `_resolveTriggeredCall` stays — it's an independent latent bug in the per-call decode path.

New regression test `testSolvencyTransientInsolvencyRepairedStillTrips` (in the renamed `LendingSolvencyPerCall.t.sol`) builds exactly your scenario: a router that drives an account insolvent on the first op and repairs it on the second within one tx. It reverts now; a tx-end-only check would have passed it.

One open question for you: the same enumerate-affected-accounts → PreTx/PostTx pattern was applied to the **per-account health** checks in pass 2 — crvUSD (`assertPostRiskIncreasingOperationHealth`) and AaveV4 Spoke (`assertAccountDataMatchesIndependentState` / liquidation health). Those carry the identical transient-then-repaired tradeoff. The genuinely pool/asset-wide tx-end conversions (Aerodrome reserves, AaveV4 Hub accounting, LlamaLend debt cap, UniV3 fees/oracle) don't, since their invariant is aggregate final state by construction. Want me to revert the crvUSD / Spoke per-account ones to per-call too for consistency, or leave them as tx-end?

fredo (approved):
Approved after the latest trigger-scoping fixes. The previously flagged example tx-end conversions have been reverted to per-call, generic lending keeps per-call solvency with the selector-prepend decode fix, and slot protection is now narrowed to per-slot storage-change triggers without reducing coverage.

PR #74, opened 2026-06-05
feat(swaps): CoW Protocol settlement example assertions

makemake-kbo:
## What

Adds a `swaps`-category example assertion bundle protecting a CoW Protocol (Gnosis Protocol v2) settlement contract against the two on-chain harms a malicious or compromised **solver** can cause that the settlement contract does *not* guard against itself.

The GPv2 settlement contract already gates `settle`/`swap` behind an allowlist of bonded solvers, enforces user signatures, limit prices, fill amounts and expiry, and forbids interactions with the vault relayer. What it intentionally leaves open — solvers may run arbitrary interactions and may dip into the contract's own accumulated buffers — is exactly what these two invariants cover. Both observe **real token movements** and neither uses a price oracle.

## Invariants

1. **`assertSolverDoesNotExtractValue`** (per `settle`/`swap` call) — *surplus theft.* The settlement must not pay any of its tokens to the solver that authored the call (solver rewards are paid out-of-band in COW). This is the on-chain footprint of surplus siphoning that the contract's signature/limit-price checks cannot catch — those only guarantee the user's signed floor, not where any excess goes.

2. **`assertBufferConserved`** (tx end) — *settlement inventory.* The settlement's accumulated buffer for each watched token may not be net-reduced across a transaction unless the full amount lands at the authorized DAO sweep recipient. Maps directly to the February-2023 DAI-buffer drain. NatSpec documents the production hardening: also register an `erc20-change` trigger per buffer token to catch drains that never call the settlement (the standing-approval class behind that incident).

## Fidelity

The minimal `IGPv2SettlementLike` interface mirrors `cowprotocol/contracts` byte-for-byte, so the derived selectors match mainnet (`settle` = `0x13d79a0b`, `swap` = `0x845a101f`), pinned in the test.

## Tests

`pcl test` — 5/5 pass: mainnet selector check, honest-settlement-pass / solver-siphon-trip, authorized-sweep-pass / unauthorized-drain-trip. The E2E uses a faithful settlement mock that reproduces the real on-chain footprint (pull sell token from user, pay buy token to receiver, retain fee buffer) plus the two malicious footprints.

## Files

- `src/protection/swaps/examples/CowSettlementInterfaces.sol`
- `src/protection/swaps/examples/CowSettlementHelpers.sol`
- `src/protection/swaps/examples/CowSettlementAssertion.sol`
- `test/protection/swaps/CowSettlementAssertion.t.sol`

### DIFF
--- a/.github/workflows/solidity-test.yml
+++ b/.github/workflows/solidity-test.yml
@@ -10,10 +10,12 @@ jobs:
     uses: phylaxsystems/actions/.github/workflows/solidity-base.yaml@main
     with:
       gas-diff-tolerance: 25
-      forge-test-args: "--no-match-path 'test/{backtesting,protection/safe}/**'"
+      # Protection-suite tests exercise the Phorge-only `cl.assertion` cheatcode, so the whole
+      # suite runs under `pcl test` in the protection-pcl job instead of base `forge test`.
+      forge-test-args: "--no-match-path 'test/{backtesting,protection}/**'"
       disable-gas-snapshot: true
 
-  safe-pcl:
+  protection-pcl:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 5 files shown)


fredo (approved):
Approved after reviewing the CoW Protocol settlement assertions and verification results. No blocking findings.

PR #75, opened 2026-06-09
feat(cap): cUSD backing-conservation and OFT lockbox assertions

makemake-kbo:
## What

Two new Cap protocol example assertions, closing the two coverage gaps the existing Cap examples (OFAC compliance + redemption gate) do not address: **infinite mint of cUSD without backing** and **cross-chain "becoming Kelp"** risk.

### 1. `CapMintBackingAssertion` — no infinite mint of cUSD
- **`assertBackingCoversSupply`** (fires on `mint`/`burn`/`redeem`): cUSD supply, valued at its **$1 face peg**, must stay covered by oracle-valued backing `Σ totalSupplies(asset) · price(asset)`. Enforced as a **non-worsening PreTx→PostTx delta** so a pre-existing depeg doesn't flag unrelated txs, while any tx that mints cUSD without adding backing (or drains backing without burning cUSD) trips it.
  - cUSD is valued at face, **not** the CapToken oracle — that oracle reports `backing / supply`, so valuing cUSD with it would make the check circular and detect nothing.
- **`assertReserveInflowAccounted`** (cumulative-inflow circuit breaker): rejects unaccounted reserve **donations / stuffing** (idle custody jumps without a matching increase in protocol accounting). This is the inflow counterpart to the existing redemption gate, which only bounds outflow.

### 2. `CapOFTLockboxBackingAssertion` — keep cross-chain cUSD honest (#"becoming Kelp")
- On the home chain, locked cUSD may leave the `OFTLockbox` (LayerZero `OFTAdapter`) **only** through a verified endpoint `lzReceive` (i.e. a settled remote burn). A drain via compromised owner/upgrade, faulty path, or stale approval — which would leave remote `L2Token` supply unbacked — reverts.

## Why
Mapped against Cap's stated concerns, the prior suite covered sanctions + bank-run outflow but **not** cUSD solvency or bridge backing. These two assertions target the external-state invariants no single protocol `require` enforces.

## Correctness basis
Valuation and accounting mirror `cap-labs-dev/cap-contracts`:
- `value = amount * price / 10**decimals`, oracle prices in 8-decimal USD (`MinterLogic`, `CapTokenAdapter`).
- `IVault.totalSupplies(asset)` is the conserved backing measure (unchanged by borrow / fractional-reserve invest).
- `ICapToken is IERC20, IVault, IMinter, IFractionalReserve` — the adopter is the combined CapToken.
- `OFTLockbox` is a LayerZero `OFTAdapter`; `L2Token` is the remote OFT.

## Testing
`FOUNDRY_PROFILE=cap pcl test` → **16/16 pass** (6 backing + 3 lockbox + 7 existing gate). Each new assertion has an honest-pass path and a tripping path; `forge fmt` clean; assertion source compiles warning-free.

## Files
- `examples/cap/src/CapMintBackingAssertion.sol` + `CapMintBackingHelpers.sol` + `CapMintBackingInterfaces.sol`
- `examples/cap/src/CapOFTLockboxBackingAssertion.sol` + `CapOFTLockboxInterfaces.sol`
- `examples/cap/test/CapMintBackingAssertion.t.sol`, `examples/cap/test/CapOFTLockboxBackingAssertion.t.sol`
- `examples/cap/README.md`

### DIFF
--- a/examples/cap/README.md
+++ b/examples/cap/README.md
@@ -33,12 +33,28 @@ FOUNDRY_PROFILE=cap pcl test
 
 ### Cross-chain backing — keep the OFT lockbox honest
 
-- `CapOFTLockboxBackingAssertion.sol` — on the home chain, locked cUSD may leave the `OFTLockbox`
-  (LayerZero `OFTAdapter`) only through a verified endpoint `lzReceive`, so remote `L2Token`
-  supply can never be left unbacked by a drained lockbox.
+- `CapOFTLockboxBackingAssertion.sol` — on the home chain, the gross outflow of locked cUSD from the
+  `OFTLockbox` (LayerZero `OFTAdapter`) must be covered by the amount actually released *inside*
+  successful endpoint-driven `lzReceive` calls. This binds the released amount and recipient to a
+  verified remote burn, so a drain cannot ride alongside an unrelated (or reverted) bridge-in and
+  leave remote `L2Token` supply unbacked.
 - `CapOFTLockboxInterfaces.sol`
 
 ### Existing examples
 
 - `CapOfacComplianceAssertion.sol` / `CapOfacComplianceInterfaces.sol` — sanctions screening.
 - `CapRedemptionGateAssertion.sol` / `CapRedemptionGateInterfaces.sol` — tiered bank-run outflow gate.
+
+## Operational contract & open decisions
+
+These examples carry two deliberate tradeoffs that Cap must own before production (also noted in
+`CapMintBackingHelpers.sol`):
+
+- **Fail-closed on read failure (decide with Cap).** Oracle/state reads revert on infra errors, so a
+  transiently down/stale/zero oracle blocks mint/burn/redeem. Defensible for a solvency invariant
+  (halt rather than allow an unbacked mint), but the alternative — fail-open on infra-level read
+  failures while keeping the solvency comparison strict — is a valid choice. Pending Cap sign-off.
+- **Static backing-asset set (must resolve before production).** `ASSET0..ASSET4` are fixed at deploy
+  via immutables. Any reserve-set change silently under-counts backing and drops inflow coverage for
+  the new asset. Before mainnet, either enforce a redeploy-on-reserve-change operational contract or
+  drive the asset set from on-chain state.
--- a/examples/cap/README.md
+++ b/examples/cap/README.md
@@ -20,6 +20,17 @@ FOUNDRY_PROFILE=cap pcl test
   peg, not the self-referential CapToken oracle).
 - `CapMintBackingInterfaces.sol`
 
+### Liquidations — debt is repaid, proceeds stay in the protocol
+
+- `CapLiquidationAssertion.sol` — deployed against the `Lender`. A liquidation that moves value
+  must strictly reduce the borrower's debt for the liquidated asset (no seizing collateral without
+  repaying debt), and the vault's claimable backing (`availableBalance = totalSupplies -
+  totalBorrows`) may not fall below its pre-call value minus the restaker interest the liquidation
+  legitimately realizes (proceeds stay in the protocol as backing rather than draining the vault).
+  Each check is scoped per `liquidate` call via PreCall/PostCall snapshots.
+- `CapLiquidationHelpers.sol` — triggered-call resolution and fork-aware reads (all in asset units).
+- `CapLiquidationInterfaces.sol`
+
 ### Cross-chain backing — keep the OFT lockbox honest
 
 - `CapOFTLockboxBackingAssertion.sol` — on the home chain, locked cUSD may leave the `OFTLockbox`
--- a/examples/cap/README.md
+++ b/examples/cap/README.md
@@ -1,16 +1,33 @@
 # cap examples
 
-Assertion examples and supporting helpers extracted from the `cap` branch.
+Assertion examples and supporting helpers for the Cap protocol (cUSD / stcUSD).
 
-## Build
+## Build & test
 
 ```sh
 FOUNDRY_PROFILE=cap forge build
+FOUNDRY_PROFILE=cap pcl test
 ```
 
 ## Files
 
-- CapOfacComplianceAssertion.sol
-- CapOfacComplianceInterfaces.sol
-- CapRedemptionGateAssertion.sol
-- CapRedemptionGateInterfaces.sol
+### Backing conservation — no infinite mint of cUSD
+
+- `CapMintBackingAssertion.sol` — cUSD supply must stay covered by oracle-valued backing
+  (`Σ totalSupplies(asset) * price(asset)`) on every mint/burn/redeem, plus a cumulative-inflow
+  circuit breaker that rejects unaccounted reserve donations/stuffing.
+- `CapMintBackingHelpers.sol` — fork-aware reads and USD valuation (cUSD valued at its $1 face
+  peg, not the self-referential CapToken oracle).
+- `CapMintBackingInterfaces.sol`
+
+### Cross-chain backing — keep the OFT lockbox honest
+
+- `CapOFTLockboxBackingAssertion.sol` — on the home chain, locked cUSD may leave the `OFTLockbox`
+  (LayerZero `OFTAdapter`) only through a verified endpoint `lzReceive`, so remote `L2Token`
+  supply can never be left unbacked by a drained lockbox.
+- `CapOFTLockboxInterfaces.sol`
+
+### Existing examples
+
+- `CapOfacComplianceAssertion.sol` / `CapOfacComplianceInterfaces.sol` — sanctions screening.
+- `CapRedemptionGateAssertion.sol` / `CapRedemptionGateInterfaces.sol` — tiered bank-run outflow gate.
--- a/examples/cap/README.md
+++ b/examples/cap/README.md
@@ -9,6 +9,11 @@ FOUNDRY_PROFILE=cap forge build
 FOUNDRY_PROFILE=cap pcl test
 ```
 
+Requires `pcl` >= 1.4.0. These assertions register cumulative-flow triggers and use the
+`getLogsForCall` precompile; older builds fail during assertion setup with
+`Fn selector not found` / `Precompile selector not found` and execute 0 tests.
+Verified with `pcl` 1.4.0 (commit `a600e1d`).
+
 ## Files
 
 ### Backing conservation — no infinite mint of cUSD
@@ -34,10 +39,12 @@ FOUNDRY_PROFILE=cap pcl test
 ### Cross-chain backing — keep the OFT lockbox honest
 
 - `CapOFTLockboxBackingAssertion.sol` — on the home chain, the gross outflow of locked cUSD from the
-  `OFTLockbox` (LayerZero `OFTAdapter`) must be covered by the amount actually released *inside*
-  successful endpoint-driven `lzReceive` calls. This binds the released amount and recipient to a
-  verified remote burn, so a drain cannot ride alongside an unrelated (or reverted) bridge-in and
-  leave remote `L2Token` supply unbacked.
+  `OFTLockbox` (LayerZero `OFTAdapter`) must be covered by endpoint-driven `lzReceive` calls, each
+  credited at `min(message-authorized amount, amount actually released inside the call)`. Capping by
+  the decoded OFT message amount binds the release to what the remote chain verifiably burned (a
+  faulty/upgraded adapter over-releasing trips); the in-call release floor gates success (a reverted
+  receive credits nothing). A drain therefore cannot ride alongside an unrelated, reverted, or
+  dust-sized bridge-in and leave remote `L2Token` supply unbacked.
 - `CapOFTLockboxInterfaces.sol`
 
 ### Existing examples
--- a/examples/cap/src/CapLiquidationAssertion.sol
+++ b/examples/cap/src/CapLiquidationAssertion.sol
@@ -1,0 +1,71 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.13;
+
+import {PhEvm} from "credible-std/PhEvm.sol";
+import {AssertionSpec} from "credible-std/SpecRecorder.sol";
+import {CapLiquidationHelpers} from "./CapLiquidationHelpers.sol";
+import {ICapLenderLike} from "./CapLiquidationInterfaces.sol";
+
+/// @title CapLiquidationAssertion
+/// @author Phylax Systems
+/// @notice Keeps Cap liquidations honest: a liquidation must actually repay the borrower's debt,
+///         and its proceeds must stay in the protocol as backing rather than draining the vault.
+/// @dev Deploy against the Cap `Lender`. A liquidation repays an agent's debt from the
+///      liquidator's funds and, in exchange, slashes the agent's restaked delegation collateral to
+///      the liquidator. Each check is scoped to a single `liquidate` call via PreCall/PostCall
+///      snapshots, so unrelated operations in the same transaction never contaminate it.
+///
+///      1. Debt is repaid (`assertLiquidationReducesDebt`): a liquidation that moves value
+///         (`amount > 0`) must strictly reduce the agent's debt for the liquidated asset. This is
+///         the "no liquidation without repaying debt" guarantee — collateral cannot be seized
+///         while the borrower's liability is left untouched. (A successful `liquidate` always has
+///         a liquidatable agent, so any positive amount repays a positive amount.)
+///
+///      2. Backing is retained (`assertLiquidationRetainsBacking`): the vault's claimable backing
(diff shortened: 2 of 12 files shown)


examples/cap/README.md
```diff
@@ -1,16 +1,44 @@
...
 
-Assertion examples and supporting helpers extracted from the `cap` branch.
+Assertion examples and supporting helpers for the Cap protocol (cUSD / stcUSD).
 
-## Build
+## Build & test
 
 ```sh
 FOUNDRY_PROFILE=cap forge build
+FOUNDRY_PROFILE=cap pcl test
```
mateo-mro, line 9: The stated Cap verification is not reproducible. After fetching submodules, `FOUNDRY_PROFILE=cap pcl test` fails during assertion setup with `Fn selector not found` and reports 0 tests executed. The new examples need a passing PCL run before this command can be documented as verification.

examples/cap/src/CapOFTLockboxBackingAssertion.sol
```diff
@@ -0,0 +1,70 @@
...
+
+        if (!_endpointReceivePresent()) {
+            revert("CapLockbox: locked cUSD released outside bridge receive");
+        }
+    }
+
+    /// @notice True when the trusted endpoint invoked `lzReceive` on the lockbox this transaction.
+    function _endpointReceivePresent() internal view returns (bool) {
+        PhEvm.CallInputs[] memory calls =
+            ph.getAllCallInputs(ph.getAssertionAdopter(), ILayerZeroReceiverLike.lzReceive.selector);
```
fredo: **Lockbox check is presence-only, not amount/recipient-bound (and ignores call success).**

`assertReleaseOnlyOnReceive` → `_endpointReceivePresent()` passes as long as *any* endpoint-driven `lzReceive` exists in the transaction. It doesn't bind the released amount or recipient to that `lzReceive`'s message, and it doesn't check the call actually succeeded (`getAllCallInputs` returns attempted calls).

Consequence: the docstring claims "the release settles a verified remote burn rather than an unauthorized drain," but co-occurrence isn't causation. An attacker who can land *any* legitimate endpoint `lzReceive` in the same tx (a dust-amount bridge-in, or even a present-but-reverted call) can drain the lockbox alongside it and this still passes. So the real "becoming Kelp" failure mode — a drain that rides in the same block/tx as routine bridge traffic — isn't caught.

To actually enforce the invariant, reconcile the *outflow amount* against the sum of successful endpoint `lzReceive` message amounts for this tx (and gate on `success`), rather than checking for mere presence. Net locked-balance delta should be `<=` the total credited by verified receives.

examples/cap/src/CapMintBackingHelpers.sol
```diff
@@ -0,0 +1,111 @@
...
+        if (asset == address(0)) return 0;
+        address adopter = ph.getAssertionAdopter();
+        uint256 supplied = _readUint(adopter, abi.encodeCall(ICapVaultLike.totalSupplies, (asset)), fork);
+        if (supplied == 0) return 0;
+        uint256 price = _readPriceUsd8(asset, fork);
+        uint256 decimalsPow = _readDecimalsPow(asset, fork);
+        value = ph.mulDivDown(supplied, price, decimalsPow);
+    }
+
+    function _readPriceUsd8(address asset, PhEvm.ForkId memory fork) private view returns (uint256 price) {
```
fredo, line 104: **Open policy decision to align with Cap: fail-closed vs. fail-open on read failure.**

These reads fail closed — `require(res.ok …)` / `require(price > 0)` here, and the generic `_readUint`. Since a reverting assertion blocks the transaction, this couples mint/burn/redeem **liveness** to oracle liveness: if Cap's price feed is transiently down, reverts, or returns a stale/zero price, *all* mint/burn/redeem get blocked — possibly exactly when the protocol is already stressed, and blocking redeems/burns can trap users.

That may well be the right call for a solvency invariant (better to halt than allow an unbacked mint), but it's a deliberate tradeoff Cap should own, not inherit by accident. We can pick either default now, but the decision itself needs to be made *with* Cap.

Two asks:
1. Capture this as a NatSpec note on the helper (and/or the cap README) so the open decision is visible in-code, e.g.:
   > POLICY (decide with Cap): reads fail closed — an oracle outage/stale price blocks mint/burn/redeem. Alternative is to treat infra-level read failures as skip-the-check while keeping the solvency condition itself fail-closed. Pending Cap sign-off.
2. Same note should flag that the alternative is fail-open-on-infra-error (oracle revert/stale ⇒ skip) while keeping the *solvency comparison* strict.

Related: the immutable `ASSET0..ASSET4` set has the same "needs a documented operational contract" flavor — it must be redeployed on any reserve-set/oracle change or it silently drifts. Worth the same kind of in-code note.

examples/cap/src/CapLiquidationAssertion.sol
```diff
@@ -0,0 +1,71 @@
...
+    ///      value and seize no collateral). Otherwise fails if the agent's debt for the liquidated
+    ///      asset did not fall across the call — i.e. collateral was seized without repaying debt.
+    function assertLiquidationReducesDebt() external view {
+        LiquidationCall memory liq = _resolveLiquidation();
+        if (liq.amount == 0) return;
+
+        uint256 debtPre = _debtAt(liq.agent, liq.asset, _preCall(liq.callStart));
+        uint256 debtPost = _debtAt(liq.agent, liq.asset, _postCall(liq.callEnd));
+
+        require(debtPost < debtPre, "CapLiquidation: debt not reduced");
```
fredo: **False-positive risk: `strictly reduces debt` can trip on an honest partial liquidation due to interest accrual.**

`debt()` is principal + accrued restaker interest, and a real `liquidate` *realizes* restaker interest (adding it to debt) before repaying. On a small/partial liquidation where the interest realized in this call is >= the principal repaid, `debtPost >= debtPre` across the call even though the liquidation was perfectly honest — and this reverts.

The `amount > 0` guard doesn't cover it: amount can be positive yet smaller than the freshly-realized interest. Consider asserting reduction of *principal* specifically, or netting out the realized interest (you already read `maxRestakerRealization`) before the comparison: e.g. require `debtPost < debtPre + realizedInterest`, or compare debt-token principal rather than total debt.

examples/cap/src/CapMintBackingAssertion.sol
```diff
@@ -0,0 +1,97 @@
...
+    ///      pre-state was already short (e.g. depeg / bad debt), the tx must not worsen it.
+    function assertBackingCoversSupply() external view {
+        int256 surplusPre = _surplusUsd8(_preTx());
+        int256 surplusPost = _surplusUsd8(_postTx());
+
+        // forge-lint: disable-next-line(unsafe-typecast) — USD-8 tolerance is far below int256 max
+        int256 tolerance = int256(_capFaceValueUsd8(_preTx()) * SOLVENCY_TOLERANCE_BPS / 10_000);
+
+        if (surplusPre >= 0) {
+            require(surplusPost >= -tolerance, "CapBacking: cUSD not fully backed");
```
fredo: **Backing floor permits eroding the entire over-collateralization buffer in one tx (hole in "no infinite mint").**

In the healthy branch the check only requires `surplusPost >= -tolerance`. So when the protocol carries surplus (say +$5M over-collateralized), a single tx can consume all of it — i.e. an unbacked mint up to the existing surplus passes without tripping. The "no infinite mint" guarantee therefore has a hole exactly the size of current over-collateralization.

If the intent is conservation rather than just a solvency floor, mirror the unhealthy branch and require non-worsening in both cases: `require(surplusPost >= surplusPre - tolerance)` regardless of sign. That still tolerates a depeg without penalizing unrelated txs, but stops a single tx from draining the buffer. If the floor-only behavior is intentional, worth a comment saying so explicitly.

examples/cap/src/CapMintBackingHelpers.sol
```diff
@@ -0,0 +1,111 @@
...
+///      (`FACE_PRICE_USD8`) rather than at the CapToken oracle, because that oracle reports
+///      cUSD as backing-over-supply and would make a backing check circular.
+abstract contract CapMintBackingHelpers is Assertion {
+    /// @dev USD price scale used by the Cap oracle (8 decimals).
+    uint256 internal constant FACE_PRICE_USD8 = 1e8;
+
+    /// @dev Gas budget for nested static reads against a fork snapshot.
+    uint64 internal constant READ_GAS = 3_000_000;
+
+    address internal immutable ORACLE;
```
fredo, line 43: **Document the asset-set drift risk in-code; must be resolved before production.**

`ASSET0..ASSET4` are immutable, set once at construction. That bakes the reserve set in at deploy time: if Cap later adds a backing asset (or migrates one), `_backingValueUsd8` silently under-counts backing -> false-positives on honest mints, and the inflow breaker never watches the new asset (silent coverage gap). The assertion has no way to know the set changed.

Please add a NatSpec note here making this explicit and a pre-production TODO, e.g.:

> DRIFT (must resolve before production): the backing asset set is fixed at deploy via immutables. Any reserve-set change (asset added/removed/migrated) silently under-counts backing and drops inflow coverage for the new asset. Either (a) enforce a redeploy-on-reserve-change operational contract, or (b) drive the asset set from on-chain state before mainnet. Do not ship to production with the static list unaddressed.

Same note belongs in examples/cap/README.md so the operational contract is discoverable outside the source.

examples/cap/src/CapOFTLockboxBackingAssertion.sol
makemake-kbo: Fixed in e132a8a. Replaced the presence-only check with amount reconciliation: `assertReleaseOnlyOnReceive` now requires gross locked-token outflow ≤ the amount actually transferred out *inside* successful endpoint-driven `lzReceive` calls (summed via `getLogsForCall` scoped to each receive call id). This binds both the released amount and recipient to a verified remote burn, and gates success implicitly — a reverted `lzReceive` commits no logs, so it credits zero. A drain riding alongside a dust (or reverted) bridge-in now inflates the outflow without matching credit and reverts. Added `testDrainRidingVerifiedReceiveTrips` covering exactly that bypass.

Note: used `getAllCallInputs` + `getLogsForCall` rather than `matchingCalls(successOnly)` because the released `pcl` build does not implement `matchingCalls` (`Precompile selector not found: 0x0dc1cc5d`).

examples/cap/src/CapLiquidationAssertion.sol
makemake-kbo: Fixed in e132a8a. `assertLiquidationReducesDebt` now nets out the realized restaker interest before comparing: `require(debtPost < debtPre + realized)`. Since `liquidate` realizes interest into the agent's debt before repaying, an honest partial liquidation whose realized interest exceeds the principal repaid no longer false-trips. The mock now realizes interest into debt as well, so `NoRepay` still trips (`debtPost == debtPre + realized`, not `<`). Added `testPartialLiquidationWithHighInterestPasses` (realized 10 > repaid 5), which fails under the old `debtPost < debtPre` check.

examples/cap/src/CapMintBackingAssertion.sol
makemake-kbo: Fixed in e132a8a. Collapsed the two branches into a single conservation check: `require(surplusPost >= surplusPre - tolerance)` regardless of sign. A single tx can no longer erode the over-collateralization buffer while ending non-negative. Honest mint/burn/redeem keep surplus flat so they still pass, and a pre-existing depeg is still tolerated (the bound is relative to `surplusPre`). Added `testUnbackedMintWithinBufferTrips` (+$100 buffer, 50 cUSD unbacked mint ends +$50 — passes a floor-only check, trips conservation).

examples/cap/src/CapMintBackingHelpers.sol
makemake-kbo: Documented in e132a8a. Added a `POLICY — fail-closed on read failure (decide with Cap)` NatSpec note on `CapMintBackingHelpers` spelling out the liveness coupling and the fail-open-on-infra-error alternative (keeping the solvency comparison itself strict), pending Cap sign-off. Mirrored in the cap README under "Operational contract & open decisions".

examples/cap/src/CapMintBackingHelpers.sol
makemake-kbo: Documented in e132a8a. Added a `DRIFT — static asset set (must resolve before production)` NatSpec note at the `ASSET0..ASSET4` immutables covering the silent under-count / inflow-coverage gap and the two resolutions (redeploy-on-reserve-change operational contract, or drive the set from on-chain state). Mirrored in the cap README.

examples/cap/README.md
makemake-kbo: Root cause is an older `pcl` build: the assertions register cumulative inflow/outflow watchers, and the lockbox fix originally also used `ph.matchingCalls`, which an older toolchain reports as `Precompile selector not found` during setup → 0 tests executed. On the current `pcl` in PATH the full suite now runs `FOUNDRY_PROFILE=cap pcl test → 25/25 pass` (e132a8a), and I removed the `matchingCalls` dependency. The remaining piece — pinning a minimum `pcl` / precompile version in the README so this is reproducible on any maintainer's toolchain — is still open, so leaving this thread unresolved for that.

examples/cap/src/CapOFTLockboxBackingAssertion.sol
```diff
@@ -0,0 +1,105 @@
...
+        PhEvm.LogQuery memory query = PhEvm.LogQuery({emitter: LOCKED_TOKEN, signature: TRANSFER_SIG});
+
+        for (uint256 i; i < calls.length; ++i) {
+            if (calls[i].caller != ENDPOINT) continue;
+
+            PhEvm.Log[] memory logs = ph.getLogsForCall(query, calls[i].id);
+            for (uint256 j; j < logs.length; ++j) {
+                // topics: [sig, from, to]; data: amount. Count only releases out of the lockbox.
+                if (logs[j].topics.length == 3 && _topicAddress(logs[j].topics[1]) == lockbox) {
+                    credited += abi.decode(logs[j].data, (uint256));
```
mateo-mro: This still credits whatever the lockbox transfers during an endpoint lzReceive, instead of the amount authorized by the verified message. If a faulty or upgraded adapter releases 500 cUSD while the message only authorizes 1 cUSD, grossOutflow and credited both include 500, so the assertion passes even though remote supply is now underbacked. Please decode the receive payload, or another canonical amount source, and cap credited by the message amount rather than by the token logs emitted during the call.

makemake-kbo: Fixed in 7c3a09e. Credit is no longer derived from the adapter's own transfer logs: `_messageAmount` decodes the verified OFT payload (`sendTo` bytes32 ++ `amountSD` uint64, converted to local units via an immutable `DECIMAL_CONVERSION_RATE`) and each receive now credits `min(messageAuthorizedAmount, amountReleasedInsideTheCall)`. The message amount — the value the remote chain verifiably burned — is the cap, so a faulty/upgraded adapter releasing 500 cUSD against a 1 cUSD message inflates gross outflow without matching credit and trips. The in-call release floor is kept only to gate success (a reverted `lzReceive` commits no logs, so it credits nothing). Mocks now use the real OFT codec, and `testFaultyAdapterOverReleaseTrips` covers exactly this bypass (adapter releases 500x the authorized amount).

examples/cap/README.md
makemake-kbo: Closed out in 7c3a09e: the README now pins the toolchain — `pcl` >= 1.4.0, with a note that older builds fail during assertion setup with `Fn selector not found` / `Precompile selector not found` and execute 0 tests (the failure you hit). Verified on pcl 1.4.0 (commit a600e1d): `FOUNDRY_PROFILE=cap pcl test` → 26/26 pass.

examples/cap/src/CapOFTLockboxBackingAssertion.sol
```diff
@@ -0,0 +1,141 @@
...
+            if (calls[i].caller != ENDPOINT) continue;
+
+            uint256 authorized = _messageAmount(calls[i].input);
+            if (authorized == 0) continue;
+
+            uint256 releasedInCall;
+            PhEvm.Log[] memory logs = ph.getLogsForCall(query, calls[i].id);
+            for (uint256 j; j < logs.length; ++j) {
+                // topics: [sig, from, to]; data: amount. Count only releases out of the lockbox.
+                if (logs[j].topics.length == 3 && _topicAddress(logs[j].topics[1]) == lockbox) {
```
mateo-mro, line 111: This caps the release amount by the verified OFT message, but it still accepts any recipient as long as the transfer leaves the lockbox inside the endpoint call. A faulty adapter can decode a message for Alice, release the same amount to Bob, and this assertion passes because only the from topic is checked. If this assertion is meant to bind the bridge release to the remote burn, it should decode sendTo and require the transfer recipient to match it, or the docs should narrow the guarantee to supply backing only.

mateo-mro (approved):

PR #76, opened 2026-06-09
feat(examples): add Lighter bridge rollup-safety assertions

makemake-kbo:
## Summary

Adds a standalone `examples/lighter` bundle of runtime Credible Layer assertions for Lighter's `ZkLighter` L1 bridge / ZK-rollup contract (`0x3B4D794a66304F130a4Db8F2551B0070dfCf5ca7`). Each invariant targets a cross-storage or rolling-window property the contract cannot enforce against itself on-chain — these bridges are run by a small validator set behind an upgradeable proxy, so an assertion observing pre/post state constrains even the privileged paths the contract trusts.

## Assertions

- **`RollupBridgeStateMachineAssertion`** — reusable base for zkSync-lineage rollup bridges:
  - **Batch & priority ordering** — `executed <= verified <= committed` for batches and the priority-request queue at transaction end.
  - **Finality non-decrease** — verified/executed counters never roll back.
  - **State-root continuity** — the executed state root moves only when a batch executes or via the proof-gated migration call.
- **`LighterBridgeAssertion`** — wires the base and adds **desert-mode integrity**: the escape hatch is irreversible and freezes the operator (no commit/verify/execute) while open.
- **`LighterOutflowCircuitBreaker`** — standalone rolling-window collateral-outflow breaker that stands down in desert mode so legitimate mass exits are never blocked.

## Tests

`FOUNDRY_PROFILE=lighter pcl test` — 19 passing:
- 11 for the state-machine + desert-mode invariants (honest + malicious path each, via `cl.assertion`).
- 8 for the breaker (pure decision logic + constructor guards).

The `watchCumulativeOutflow` trigger is not fired by local `pcl test`, so the breaker's breach path is validated through its pure `_breakerTrips` decision function (and exercised by the executor in production). This is documented in the README and NatSpec, consistent with the in-tree `ERC4626CumulativeOutflowAssertion`.

## Notes

- Self-contained under `examples/lighter/`; no changes to the core suite. Adds a `[profile.lighter]` to `foundry.toml`.
- `IZkLighterLike` assumes same-named public getters; a prominent caveat says to verify selectors/storage layout against the deployment (or switch to `loadStateAt` slot reads).


examples/lighter/test/LighterBridgeAssertion.t.sol
```diff
@@ -0,0 +1,228 @@
...
+        bridge = new MockZkLighter();
+        // Baseline consistent state machine: executed <= verified <= committed, open covers
+        // committed-but-unexecuted priority requests.
+        bridge.seed(10, 8, 6, 10, 8, 6, 4, ROOT_A, false);
+    }
+
+    function _arm(bytes4 fnSelector) internal {
+        bytes memory createData =
+            abi.encodePacked(type(LighterBridgeAssertion).creationCode, abi.encode(address(bridge)));
+        cl.assertion(address(bridge), createData, fnSelector);
```
mateo-mro, line 153: FOUNDRY_PROFILE=lighter pcl test currently fails before the first test runs. cl.assertion reverts while building the assertion state with Fn selector not found, so the 19 passing tests claimed in the PR are not reproducible and this profile cannot be used as verification until the trigger setup is fixed.

test/protection/lending/LendingSolvencyPerCall.t.sol
```diff
@@ -0,0 +1,188 @@
...
+    address internal borrower = makeAddr("borrower");
+
+    function setUp() public {
+        pool = new MockLendingPool();
+    }
+
+    function _arm() internal {
+        bytes memory createData =
+            abi.encodePacked(type(TestLendingSolvencyAssertion).creationCode, abi.encode(address(pool)));
+        cl.assertion(address(pool), createData, LendingBaseAssertion.assertOperationSafety.selector);
```
makemake-kbo, line 150: ci passes

examples/lighter/test/LighterBridgeAssertion.t.sol
makemake-kbo: this does indeed run, would be good to get more artifacts from the agent if this does not indeed hold

mateo-mro (approved):

PR #77, opened 2026-06-11
Fluid assertions

makemake-kbo:
## Summary

- Count Fluid-recognized mainnet Zircuit custody for weETH/weETHs in the Liquidity solvency assertion.
- Make the Fluid fToken share-price assertion compare a fixed-share `convertToAssets(1e12)` sample instead of `totalAssets / totalSupply`.
- Tighten the Fluid flow breaker by filtering to successful CALLs and rejecting unsupported native/external-custody token configs.
- Update Fluid example tests and docs for the corrected assumptions.

## Validation

- `FOUNDRY_PROFILE=fluid pcl test`

### DIFF
--- a/examples/fluid/README.md
+++ b/examples/fluid/README.md
@@ -24,6 +24,9 @@ FOUNDRY_PROFILE=fluid forge build
 FOUNDRY_PROFILE=fluid pcl test
 ```
 
+This suite runs on every PR via the `examples-pcl` matrix job in
+`.github/workflows/solidity-test.yml`, so a regression in these assertions fails CI.
+
 ## Assertions
 
 ### `FluidLiquiditySolvencyAssertion` — Liquidity Layer (business logic)
--- a/examples/fluid/README.md
+++ b/examples/fluid/README.md
@@ -1,0 +1,85 @@
+# Fluid (Instadapp) examples
+
+Credible Layer assertion examples for [Fluid](https://instadapp.io/product/fluid)
+([fluid-contracts-public](https://github.com/Instadapp/fluid-contracts-public)).
+
+Fluid is built around a single **Liquidity Layer** that custodies every token for every protocol on
+top of it (the lending **fTokens** and the borrow/collateral **Vaults**). The assertions here cover
+two angles the user asked for:
+
+1. **Business-logic invariants** — the core accounting facts the protocol must never violate.
+2. **Assertions that let Fluid run looser limits / riskier assets safely** — the protocol keeps its
+   generous, capital-efficient on-chain limits and aggressive risk parameters, while the assertion
+   enforces the tighter boundary that actually keeps the system solvent and liquidatable.
+
+## Build
+
+```sh
+FOUNDRY_PROFILE=fluid forge build
+```
+
+## Test
+
+```sh
+FOUNDRY_PROFILE=fluid pcl test
+```
+
+## Assertions
+
+### `FluidLiquiditySolvencyAssertion` — Liquidity Layer (business logic)
+
+Installed on the Liquidity Layer singleton. Two transaction-end invariants, read straight from the
+singleton's packed per-token storage (no resolver trust, no interest-accrual reimplementation):
+
+- **Custody covers net supply** — `balanceOf(Liquidity, token) + totalBorrow >= totalSupply` for
+  every monitored token, i.e. accrued protocol revenue stays non-negative. This is the protocol-wide
+  insolvency condition; the per-operation `require`s never state it against the real ERC20 balance.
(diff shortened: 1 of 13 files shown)


examples/fluid/README.md
```diff
@@ -0,0 +1,90 @@
...
+## Build
+
+```sh
+FOUNDRY_PROFILE=fluid forge build
+```
+
+## Test
+
+```sh
+FOUNDRY_PROFILE=fluid pcl test
```
mateo-mro, line 24: Can we wire this profile into CI as well? The green protection PCL check does not appear to run the new fluid tests, so this example can regress without the PR checks catching it.

examples/fluid/src/FluidLiquidityHelpers.sol
```diff
@@ -0,0 +1,154 @@
...
+        view
+        returns (uint256 supplyExchangePrice, uint256 borrowExchangePrice)
+    {
+        uint256 epac = _readSlot(_exchangePricesSlot(token), fork);
+        supplyExchangePrice = (epac >> BITS_SUPPLY_EXCHANGE_PRICE) & MASK_64;
+        borrowExchangePrice = (epac >> BITS_BORROW_EXCHANGE_PRICE) & MASK_64;
+    }
+
+    function _exchangePricesSlot(address token) internal pure returns (bytes32) {
+        return keccak256(abi.encode(token, SLOT_EXCHANGE_PRICES_AND_CONFIG));
```
mateo-mro: These slot helpers run once per token on every tx end check. Since both inputs are fixed width, we can avoid the `abi.encode` allocation here and hash the two words from scratch memory. Forge build is also flagging this, and it keeps the assertion cheaper when the watched token list grows.

examples/fluid/README.md
makemake-kbo: Done in bc914f7. Added an `examples-pcl` matrix job to `.github/workflows/solidity-test.yml` that runs `FOUNDRY_PROFILE=fluid pcl test` on every PR (same pcl install as the protection job). It is a matrix so future example profiles just get appended to the list. The README `## Test` section now notes the suite is CI-gated. I scoped it to fluid rather than all ~20 example profiles to avoid pulling unrelated examples (some of which may need fork RPC) into this PR.

examples/fluid/src/FluidLiquidityHelpers.sol
makemake-kbo: Done in bc914f7. `_exchangePricesSlot`/`_totalAmountsSlot` now hash the two fixed-width words straight from EVM scratch space (`mstore(0x00, token); mstore(0x20, baseSlot); keccak256(0x00, 0x40)`) instead of allocating an `abi.encode` buffer — identical slot, no memory growth, and the `asm-keccak256` lint is cleared. Verified the solvency suite (which reads storage through these slots) still passes under `pcl test`.

examples/fluid/src/FluidLiquidityFlowBreakerAssertion.sol
```diff
@@ -0,0 +1,112 @@
...
+        uint256 length = tokens.length;
+        for (uint256 i; i < length; ++i) {
+            if (tokens[i] == token) return true;
+        }
+        return false;
+    }
+
+    function _successfulOperateCalls() internal pure returns (PhEvm.CallFilter memory filter) {
+        filter.callType = 1; // CALL only: ignore failed probes, staticcalls and delegatecalls.
+        filter.successOnly = true;
```
mateo-mro: Can we set `maxDepth` here to `type(uint32).max`, or use `_successOnlyFilter()` and then override `callType` to `1`? The assertion says it scans successful `operate` calls to the singleton, and the shared helper treats that as any depth. Leaving the depth fields at their zero defaults makes this depend on precompile defaults, and the test only pins `callType` and `successOnly`, so a router or proxy path could be missed if zero means depth zero.

makemake-kbo: Good catch — fixed in e27fdb8 using the `_successOnlyFilter()` reuse you suggested. `_successfulOperateCalls()` now starts from the shared filter (`minDepth 0`, `maxDepth type(uint32).max`, any depth) and narrows `callType` to `1`, so router/proxy-routed `operate` calls at depth > 1 are scanned instead of relying on the precompile's interpretation of a zero `maxDepth`. Extended `testWarningTierUsesSuccessfulCallFilter` to also pin `minDepth == 0`, `maxDepth == type(uint32).max`, and `topLevelOnly == false`.

mateo-mro (approved):

PR #78, opened 2026-06-12
fix(perpetual): prepend selector in PerpetualBaseAssertion._resolveTriggeredCall

makemake-kbo:
## Problem

`ph.getAllCallInputs(...).input` returns calldata args **without** the 4-byte selector, but `decodeOperation` consumers expect selector-prefixed calldata (they use `triggered.input[4:]`). `LendingBaseAssertion._resolveTriggeredCall` re-prepends the selector (added in #72), but **`PerpetualBaseAssertion._resolveTriggeredCall` was missed** — it forwarded `calls[i].input` raw.

So in production the Denaria suite's `triggered.input[4:]` stripped four bytes of the real first argument and decoded a **corrupted account**. The decode unit tests didn't catch it because they feed `decodeOperation` selector-prefixed `abi.encodeCall(...)` input directly — masking the test-harness-vs-production divergence flagged in #72.

## Fix

- **`PerpetualBaseAssertion.sol`** — prepend the selector in `_resolveTriggeredCall` (`bytes.concat(context.selector, calls[i].input)`), mirroring `LendingBaseAssertion`. This makes the perpetual `TriggeredCall.input` contract match the lending one. **No Denaria source change needed** — its `[4:]` was correct against the intended contract.
- **`examples/royco/src/RoycoHelpers.sol`** — fix a separate double-strip in `_resolveTriggeredKernelRedeemCall`: `ph.matchingCalls(...).input` also has the selector already stripped, so decode it directly. The `callinputAt`-based decoders there correctly keep `_stripSelector`.
- **`test/protection/perpetual/PerpetualOperationSafetyPerCall.t.sol`** (new) — minimal combined-suite assertion (kept under the 24 KB limit) that exercises the real `getAllCallInputs → prepend → decode` path through `cl.assertion`. Verified it **fails without the prepend** and passes with it.

## Not changed (and why)

**Tydro is not buggy.** It flows through `LendingBaseAssertion` (which already prepends), so its `triggered.input[4:]` is correct — identical to the `AaveV3LikeOperationSafety` reference. Dropping `[4:]` there would have broken it.

## Verification

- `pcl test --match-path 'test/protection/**'` → **83 passed, 0 failed** (incl. 3 new tests + existing lending per-call suite).
- Royco example compiles cleanly.
- One caveat: `ph.matchingCalls` isn't implemented in the local SDK / `pcl test` yet, so the Royco fix can't be exercised E2E locally — it relies on the `TriggerCall` struct's separate `selector` field and consistency with the rest of the `getAllCallInputs` family.

Refs: surfaced during review of #75; same selector-stripping class as #72.

### DIFF
--- a/src/protection/perpetual/PerpetualBaseAssertion.sol
+++ b/src/protection/perpetual/PerpetualBaseAssertion.sol
@@ -496,11 +496,13 @@ abstract contract PerpetualBaseAssertion is Assertion {
 
         for (uint256 i; i < calls.length; ++i) {
             if (calls[i].id == context.callStart) {
+                // `getAllCallInputs` returns calldata args WITHOUT the 4-byte selector (it is the
+                // query key), but `decodeOperation` expects selector-prefixed calldata. Prepend it.
                 return IPerpetualProtectionSuite.TriggeredCall({
                     selector: context.selector,
                     caller: calls[i].caller,
                     target: calls[i].target_address,
-                    input: calls[i].input,
+                    input: bytes.concat(context.selector, calls[i].input),
                     callStart: context.callStart,
                     callEnd: context.callEnd
                 });
--- a/test/protection/perpetual/PerpetualOperationSafetyPerCall.t.sol
+++ b/test/protection/perpetual/PerpetualOperationSafetyPerCall.t.sol
@@ -1,0 +1,210 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.13;
+
+import {Test} from "forge-std/Test.sol";
+
+import {CredibleTest} from "../../../src/CredibleTest.sol";
+import {PhEvm} from "../../../src/PhEvm.sol";
+import {AssertionSpec} from "../../../src/SpecRecorder.sol";
+import {IPerpetualProtectionSuite} from "../../../src/protection/perpetual/IPerpetualProtectionSuite.sol";
+import {PerpetualBaseAssertion} from "../../../src/protection/perpetual/PerpetualBaseAssertion.sol";
+
+interface ITestPerp {
+    function op(address account) external;
+    function equityOf(address account) external view returns (int256);
+}
+
+/// @notice Minimal perp: a single risk-reducing `op(address)` plus a per-account equity getter.
+contract MockPerp is ITestPerp {
+    mapping(address => int256) internal equity;
+    mapping(address => int256) internal pending;
+
+    function setEquity(address account, int256 value) external {
+        equity[account] = value;
+    }
+
+    function setPending(address account, int256 value) external {
+        pending[account] = value;
+    }
+
+    function op(address account) external override {
+        equity[account] = pending[account];
+    }
+
+    function equityOf(address account) external view override returns (int256) {
+        return equity[account];
+    }
+}
+
+/// @notice Minimal combined suite + perpetual base, returning itself as the suite. Implements the
+///         `IPerpetualProtectionSuite` surface directly (rather than via `PerpetualProtectionSuiteBase`)
+///         to stay under the contract-size limit, so the assertion actually deploys and exercises the
+///         generic per-call decode + post-mutation risk path — including the getAllCallInputs
+///         selector-prepend in `PerpetualBaseAssertion._resolveTriggeredCall`. Without that prepend,
+///         `decodeOperation`'s `triggered.input[4:]` strips four bytes of the real first argument and
+///         reads a corrupted account, so an honest op would no longer decode `borrower`.
+contract TestPerpRiskAssertion is PerpetualBaseAssertion, IPerpetualProtectionSuite {
+    address internal immutable PERP;
+
+    constructor(address perp_) {
+        PERP = perp_;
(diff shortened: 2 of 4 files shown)


test/protection/perpetual/PerpetualOperationSafetyPerCall.t.sol
```diff
@@ -0,0 +1,210 @@
...
+    MockPerp internal perp;
+    address internal borrower = makeAddr("borrower");
+
+    function setUp() public {
+        perp = new MockPerp();
+    }
+
+    function _arm() internal {
+        bytes memory createData = abi.encodePacked(type(TestPerpRiskAssertion).creationCode, abi.encode(address(perp)));
+        cl.assertion(address(perp), createData, PerpetualBaseAssertion.assertOperationSafety.selector);
```
mateo-mro, line 182: This test does not arm today. The harness reports Fn selector not found at cl.assertion because the selector is taken from PerpetualBaseAssertion rather than the deployed TestPerpRiskAssertion type. That means the new production path coverage is not actually running. Please use TestPerpRiskAssertion.assertOperationSafety.selector and rerun the focused PCL test.

src/protection/perpetual/PerpetualBaseAssertion.sol
```diff
@@ -496,11 +496,13 @@ abstract contract PerpetualBaseAssertion is Assertion {
 
         for (uint256 i; i < calls.length; ++i) {
             if (calls[i].id == context.callStart) {
+                // `getAllCallInputs` returns calldata args WITHOUT the 4-byte selector (it is the
```
mateo-mro, line 499: This comment fixes one reader, but the public PhEvm contract still leaves the input shape ambiguous. This fix depends on getAllCallInputs returning ABI arguments without the selector, while callinputAt returns raw selector prefixed calldata. Since this selector stripping bug has now repeated across base assertions and examples, please document the query input shape in PhEvm too so future consumers do not guess and strip or prepend incorrectly.

mateo-mro (dismissed):

test/protection/perpetual/PerpetualOperationSafetyPerCall.t.sol
makemake-kbo: Good catch on the intent, but the literal change does not compile: `assertOperationSafety()` is declared on the abstract `PerpetualBaseAssertion` and inherited (not redeclared) by `TestPerpRiskAssertion`, so `TestPerpRiskAssertion.assertOperationSafety.selector` errors with "Member not found". The selector value is identical either way (it is derived from the signature, not the referencing type), and the harness resolves it against the deployed bytecode — this mirrors the sibling lending suite which arms via `LendingBaseAssertion.assertOperationSafety.selector` (test/protection/lending/LendingSolvencyPerCall.t.sol:150) and runs fine. The "Fn selector not found" you saw is the sandbox cheatcode limitation `0x85f59e0e` that phylax-ci also flagged, not a real arming failure. I added a comment at the call site explaining this in 1c39cc7.

src/protection/perpetual/PerpetualBaseAssertion.sol
makemake-kbo: Done in 1c39cc7. Documented the query input shape directly in `PhEvm.sol`: `CallInputs.input` (returned by `getAllCallInputs`/`get*CallInputs`) and `TriggerCall.input` (returned by `matchingCalls`) are now documented as ABI args WITHOUT the 4-byte selector, with explicit guidance to prepend via `bytes.concat(selector, input)` and to NOT slice `input[4:]`. `callinputAt` already documents the contrasting selector-prefixed shape. This should stop the strip/prepend guessing across future consumers.

test/protection/perpetual/PerpetualOperationSafetyPerCall.t.sol
```diff
@@ -0,0 +1,222 @@
...
+        return operation.account != address(0) && operation.reducesAccountSafety;
+    }
+
+    /// @dev Reads the post-call equity for the decoded account and gates on `equity >= 0`.
+    function getPostMutationSnapshot(
+        TriggeredCall calldata,
+        OperationContext calldata operation,
+        PhEvm.ForkId calldata fork
+    ) external view override returns (AccountSnapshot memory snapshot) {
+        int256 equity = abi.decode(
```
mateo-mro, line 91: The focused fmt check rewrites this new test file. Can we run forge fmt here before merging so the protection profile stays clean?

makemake-kbo: This file is already formatted on the current tip (`1c39cc7`) — it was run through `forge fmt` in the dedicated style commit `89ca58c` and again in `1c39cc7`. I just re-verified: `forge fmt --check` exits 0 under both the default and `ci` profiles, and a `forge fmt` write produces no diff. There is no repo `[fmt]` config, so defaults apply. If your focused check still rewrites it, it is likely a different `forge`/`foundry` version — happy to match a specific one if you can share it.

mateo-mro (approved):

PR #79, opened 2026-06-15
feat(lido): add stETH vault flow assertion suite

makemake-kbo:
## Lido stETH vault flow assertion suite

Generic Credible Layer assertions for **any vault that custodies stETH/wstETH and deploys it into an Aave-style market** (Veda BoringVault / GGV, Mellow flexible-vaults / EarnETH, Lido V3 stVaults, or comparable loopers). Everything is expressed at transaction boundaries, configured with addresses and thresholds — no vault-stack-specific coupling.

### Assertions

| Contract | What it enforces |
|---|---|
| `LidoStEthExitBufferAssertion` | **Withdrawable-stETH buffer floor** on every transaction (idle stETH-equivalent stays above the larger of an absolute minimum and a fraction of total exposure → the vault is never fully deployed) plus a **rolling-window outflow circuit breaker** on stETH/wstETH. The breaker runs in the block builder, so a burst drain is never included even against a fully compromised signer pipeline. |
| `LidoStEthVaultRiskAssertion` | Reduce-only regime under unhealthy / off-peg / illiquid conditions, exit-liquidity and collateral-withdrawability guards, and a health-factor + collateral-ratio position envelope. |
| `LidoStEthVaultPegAssertion` | Supply-change depeg gate (any mint/burn reverts while stETH/ETH is outside the peg band) and wstETH rate integrity. |
| `LidoStEthVaultNavAssertion` | Reported share rate checked against NAV recomputed from on-chain state, in both directions. |

Shared `LidoVaultHelpers` / `LidoVaultInterfaces` back all four.

### Honest boundaries

- **Requestable, not claimable.** On-chain we can guarantee stETH stays *requestable* to Lido's withdrawal queue at any time; instant claimability is bounded by the validator exit queue and the 1,000 ETH/request cap, which no per-vault check controls.
- **Outflow breaker is destination-blind in v1** — it rate-limits all stETH/wstETH leaving the vault; a destination-aware variant (exempt transfers to Lido's `WithdrawalQueueERC721`) is the named next step.
- No protection against market-price loss; third-party Aave liquidations are out of reach; Lido core is untouched.

### Build

```sh
FOUNDRY_PROFILE=lido forge build
```

Builds clean; `forge fmt --check` passes.

Closes ENG-3428.

### DIFF
--- a/examples/lido/README.md
+++ b/examples/lido/README.md
@@ -10,11 +10,18 @@ without binding to any one stack's function selectors. Every invariant is expres
 boundaries (`registerTxEndTrigger` + pre/post-tx fork reads) or as a rolling cumulative-flow
 trigger, so the suite only needs addresses and thresholds to deploy against a concrete vault.
 
-## Build
+## Build & test
 
 ```sh
 FOUNDRY_PROFILE=lido forge build
+FOUNDRY_PROFILE=lido pcl test
 ```
+
+Requires `pcl` >= 1.4.0. The suite registers `registerTxEndTrigger` assertions and reads protocol
+state through `ph.staticcallAt` at pre/post-tx forks; every invariant below ships with pass-and-trip
+`CredibleTest` coverage under `test/` (mock pool / oracle / feed / rate-source / token knobs that
+trip one invariant at a time). The `watchCumulativeOutflow` breaker is executor-driven and not
+simulated by local `pcl test`, so its hard-revert decision is exercised directly.
 
 ## The headline ask: keep stETH withdrawable back to Lido
 
@@ -87,7 +94,7 @@ For the full "never fully deployed and never drained" guarantee, configure both.
 
 | Invariant | Failure point covered |
 |---|---|
-| Reduce-only regime: when the pre-tx health factor is below the comfort band, stETH trades off peg, the rate source can't report a rate, or the collateral reserve is no longer liquid enough for the vault to exit, a transaction may not grow debt or lower the health factor — and under market-condition triggers (shaky pricing, illiquid collateral) it may not grow the supplied position either | no new allocations into unhealthy positions; reduce-only when oracles are shaky |
+| Reduce-only regime: when the pre-tx health factor is below the comfort band, stETH trades off peg (or the stETH/ETH feed is stale, incomplete, or unreadable — fail-closed), the rate source can't report a rate, or the collateral reserve is no longer liquid enough for the vault to exit, a transaction may not grow debt or lower the health factor — and under market-condition triggers (shaky pricing, illiquid collateral) it may not grow the supplied position either | no new allocations into unhealthy positions; reduce-only when oracles are shaky |
 | Exit-liquidity guard: a transaction that grows debt must leave the borrowed reserve holding ≥ a configured multiple of the vault's debt | underlying market health (the vault can always repay and unwind) |
 | Collateral withdrawability guard: a transaction that deepens the supplied position must leave the collateral reserve holding ≥ a configured fraction of the vault's supplied collateral in un-borrowed form | the deployed leg stays withdrawable at execution time — distinct from the vault-held buffer floor above |
 | Position envelope: the health factor ends every transaction above a hard floor, may only decline while staying inside the comfort band, and the raw collateral/debt ratio holds a minimum | health factor must not degrade; collateral ratio maintained |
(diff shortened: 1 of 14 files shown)


examples/lido/src/LidoStEthVaultRiskAssertion.sol
```diff
@@ -0,0 +1,265 @@
...
+        uint256 minCollateralLiquidityBps;
+    }
+
+    constructor(RiskConfig memory config) {
+        require(config.vault != address(0), "LidoVault: zero vault");
+        require(config.aavePool != address(0), "LidoVault: zero pool");
+        require(config.minHealthFactor >= 1e18, "LidoVault: floor below liquidation");
+        require(config.reduceOnlyHealthFactor >= config.minHealthFactor, "LidoVault: band below floor");
+        require(
+            config.minExitLiquidityBps == 0
```
mateo-mro: The scoped formatter check fails on this block. The PR body says formatting passes, but this changed file still needs forge fmt.

examples/lido/README.md
```diff
@@ -0,0 +1,184 @@
...
+- LidoStEthExitBufferAssertion.sol
+- LidoStEthVaultRiskAssertion.sol
+- LidoStEthVaultPegAssertion.sol
+- LidoStEthVaultNavAssertion.sol
+- LidoVaultHelpers.sol
+- LidoVaultInterfaces.sol
+
+## Next steps
+
+- `CredibleTest` E2E coverage (mock pool/rate-source/feed/token knobs that trip one invariant at a time)
```
mateo-mro: I would not leave this as a next step in the PR. These are executable assertions with threshold and oracle edge cases, so the examples should land with CredibleTest coverage for each invariant.

examples/lido/src/LidoVaultHelpers.sol
```diff
@@ -0,0 +1,75 @@
...
+            return false;
+        }
+
+        PhEvm.StaticCallResult memory result =
+            ph.staticcallAt(feed, abi.encodeCall(IChainlinkFeedLike.latestRoundData, ()), FORK_VIEW_GAS, fork);
+        if (!result.ok) {
+            return true;
+        }
+
+        (, int256 answer,,,) = abi.decode(result.data, (uint80, int256, uint256, uint256, uint80));
```
mateo-mro: latestRoundData also returns updatedAt and answeredInRound. This should fail closed on stale or incomplete rounds, probably with a configured max age, otherwise an old peg price can keep the risk and supply change gates open during an oracle outage.

mateo-mro (changes requested):

examples/lido/src/LidoStEthVaultRiskAssertion.sol
makemake-kbo: Fixed in 29c0f58. `forge fmt --check` passes on this file with my local toolchain (forge 1.5.1 / solc 0.8.30), so the failure was a forge-fmt version disagreeing on this deeply-nested boolean. Rather than chase a version, I removed the version-sensitive construct: the `minExitLiquidityBps` / `minCollateralLiquidityBps` validations are now `if`-guarded, one short `require` per token (no line wrapping for any formatter to disagree on), which also yields more precise errors (`zero borrowed asset`, `zero collateral supply token`, etc.).

examples/lido/README.md
makemake-kbo: Done in 29c0f58 — dropped the next-step and landed coverage. Added `examples/lido/test/` with 59 passing CredibleTest tests (one pass + one trip per invariant) plus the oracle/threshold edge cases: depeg both directions, stale / incomplete / carried-over feed rounds, wstETH rate-decrease & provider-desync, buffer floor (relative + absolute + wstETH-at-Lido-rate), exit-liquidity, collateral-withdrawability, the full health-factor envelope, and NAV above/below tolerance + insolvency. Shared mocks live in `test/LidoMocks.sol`. Run with `FOUNDRY_PROFILE=lido pcl test`. The `watchCumulativeOutflow` breaker isn't simulated by local `pcl test` (executor-driven), so its hard-revert is exercised directly and noted, consistent with the Lighter example.

examples/lido/src/LidoVaultHelpers.sol
makemake-kbo: Fixed in 29c0f58. `_isStEthDepeggedAt` now decodes `updatedAt` and `answeredInRound` and fails closed on: a non-positive answer, an incomplete round (`updatedAt == 0`), a carried-over answer (`answeredInRound < roundId`), and an answer older than a configurable `maxFeedStalenessSecs` (zero disables only the age bound; the round-integrity checks always apply). I threaded `maxFeedStalenessSecs` through both consumers (the risk `RiskConfig` and the peg constructor), so an oracle outage can no longer hold the risk-allocation or share mint/burn gates open on a stale on-peg price. Covered by `testMintWhileFeedStaleTrips` / `testMintWhileRoundIncompleteTrips` / `testMintWhileRoundStaleTrips`, and documented in the README (suggested ~25h heartbeat+margin).

mateo-mro (approved):

PR #81, opened 2026-06-18
feat(mellow): flexible-vaults compromised-curator assertion suite

makemake-kbo:
## Summary

Adds a Credible Layer example assertion suite for **Mellow flexible-vaults** (the Core Vaults line: `Vault` + `Subvault`s + `RiskManager` + `Oracle` + redeem/deposit queues), under `examples/mellow/`.

The suite is scoped deliberately to **one threat model**: bound the blast radius of a **compromised curator key** (or a buggy/rogue adapter), using on-chain, per-transaction invariants the protocol's own `require`s do not already express. Mellow's hot path is permissioned by design and collateral value lives off-chain (restaking positions reconciled via oracle report), so generic vault assertions add little here — the value-add is insurance against a stolen privileged key. Every threshold is a blast-radius cap calibrated **above** the largest legitimate flow, not a tight detector.

## What's included

Four focused assertions, each adopted by a different contract because each defends a different curator-power surface:

- **`MellowVaultOutflowAssertion`** (adopter: `Vault`) — rolling-window `watchCumulativeOutflow` breaker on deposit-asset custody. Asset-flow based rather than role based, so it survives a key that passes every role check.
- **`MellowOracleReportGuardAssertion`** (adopter: `Oracle`) — immutable per-`submitReports` cap on price-per-share (`priceD18`) drift. Not a restatement of the protocol's own guard, whose `securityParams` cap is mutable and can be widened by a `SET_SECURITY_PARAMS_ROLE` key; skips bootstrap and protocol-flagged-suspicious reports.
- **`MellowRiskManagerBalanceAssertion`** (adopter: `RiskManager`) — bounds the magnitude of a single `modifyVaultBalance` / `modifySubvaultBalance` correction in **both** directions (the protocol bounds only the positive direction), relative to the pre-call balance with an absolute floor.
- **`MellowSubvaultAllocationAssertion`** (adopter: a `Subvault`) — blocks growing a supplied position in a lending market that lacks enough immediately-withdrawable liquidity to exit. Complements the per-subvault `Verifier`, which authorizes the *call* but never the *state of the market*.

Plus shared fork reads / calldata decode / signed-int math (`MellowCuratorHelpers`), minimal `Oracle`/`RiskManager` interfaces (`MellowCuratorInterfaces`), and shared mocks. Full rationale, calibration starting points, and scope/limitations are in `examples/mellow/README.md`.

## Testing

- `FOUNDRY_PROFILE=mellow pcl test` — 33 tests, pass-and-trip `CredibleTest` coverage per assertion (one mock knob per failure mode, so each test trips exactly one invariant and revert reason).
- Wired into CI via a new `examples-pcl` matrix job in `.github/workflows/solidity-test.yml`, so the suite is gated on every PR.

Linear: [ENG-3475](https://linear.app/phylaxsystems/issue/ENG-3475)


foundry.toml
```diff
@@ -131,6 +131,14 @@ cache_path = "examples/lighter/cache"
 libs = ["lib"]
 remappings = ["credible-std/=src/"]
 
+[profile.mellow]
```
mateo-mro, line 142: Can we add the mellow profile to CI before this lands? The current protection PCL job only ran the shared protection tests, so these new example assertions can regress without a required check catching it.

examples/mellow/src/MellowCuratorHelpers.sol
```diff
@@ -0,0 +1,83 @@
...
+///      `triggers()`, and the assertion functions. Everything is snapshot-scoped (PreCall/PostCall/
+///      PostTx) — there is no assertion-authored persistent state.
+abstract contract MellowCuratorHelpers is Assertion {
+    /// @notice Reads `RiskManager.vaultState().balance` (approximate shares) at a snapshot fork.
+    function _readVaultBalanceShares(address riskManager, PhEvm.ForkId memory fork)
+        internal
+        view
+        returns (int256 balance)
+    {
+        (balance,) = abi.decode(
```
mateo-mro, line 23: This file is not formatted under the mellow profile yet. The local mellow format check rewrites this decode expression, so the PR will start from a dirty formatting baseline unless we run fmt on the new example.

mateo-mro (dismissed):

foundry.toml
makemake-kbo: Done in b5f8ef5. Added an `examples-pcl` matrix job to `.github/workflows/solidity-test.yml` that runs `FOUNDRY_PROFILE=mellow pcl test` on every PR (33 tests, all passing). Mirrors the job structure used for the fluid example, so when both land the matrix just resolves to `[fluid, mellow]`. The matrix is extensible for future example profiles.

examples/mellow/src/MellowCuratorHelpers.sol
makemake-kbo: I cannot reproduce a formatting diff on the current tip (`b5f8ef5`): `forge fmt --check examples/mellow/src examples/mellow/test` exits 0 under both the default and `mellow` profiles, and a `forge fmt` write (under `FOUNDRY_PROFILE=mellow`) produces no diff on `MellowCuratorHelpers.sol`. There is no repo `[fmt]` config, so defaults apply. If your local check still rewrites the decode expression, it is likely a different `forge`/`foundry` version — let me know which and I will match it.

examples/README.md
```diff
@@ -21,6 +21,7 @@ FOUNDRY_PROFILE=<protocol> forge build
 | `denaria/` | `all-protection-suites` | Denaria perpetual operation safety |
 | `euler/` | `eulerv2` | EVault, circuit breaker, sandwich detection |
 | `lido/` | `lido` | Lido stETH vaults (generic): withdrawable-stETH buffer floor + outflow breaker, position risk regime, depeg gates, rate-vs-NAV |
+| `mellow/` | `mellow` | Mellow flexible-vaults compromised-curator suite: vault outflow breaker, immutable oracle-report drift cap, RiskManager balance-correction containment, vault config/proxy lock, subvault allocation-health (no allocating into illiquid markets) |
```
mateo-mro: This row still advertises `vault config/proxy lock`, but the assertion and tests were removed in `bfd3755`. The top level examples index will point readers at a non existent Mellow check. Can we drop that phrase here so the index matches the four assertions left in `examples/mellow/README.md`?

makemake-kbo: Done in 46e2e8b. Dropped `vault config/proxy lock` from the Mellow row so the index matches the four assertions in `examples/mellow/README.md`. (Also rebased the branch onto master, resolving the `examples-pcl` matrix to `[fluid, mellow]` as anticipated.)

mateo-mro (dismissed):

mateo-mro (approved):

PR #82, opened 2026-06-19
feat(lido): EasyTrack flash-loan governance assertion

makemake-kbo:
## Summary

Adds a Credible Layer assertion that blocks **flash-loaned governance power** from being exercised against a Lido governance contract — the motivating case being **EasyTrack**, raised by Lido on a recent call.

EasyTrack is optimistic governance: whitelisted factories create motions that execute after an objection window unless LDO holders object past a threshold. The objection weight is read from the governance token at the motion's **start block** rather than the block before. Because a same-block balance read returns the live balance, an attacker can, in one transaction, **flash-loan LDO, exercise the balance-weighted action, and repay the loan** — borrowed voting power that was never durably held still counts. EasyTrack is not upgradable, so the snapshot can't simply be moved to `block.number - 1`; an external execution gate is the agile patch.

The assertion re-creates the "snapshot a block earlier" property from the outside, so the governance contract needs no upgrade.

## Invariant

Voting power exercised against the protected selectors must be power the actor *already held at the start of the transaction* — not power acquired within the same transaction.

## Two detection layers (both independently armable)

1. **`assertNoFlashLoanedVotingPower`** (primary, precise) — for every protected call, the caller's LDO balance at `PreCall` must not exceed its balance at `PreTx` by more than `maxIntraTxAcquired`. Reads at exactly the point the governance snapshot would: a flash loan repaid before the call reads back to baseline and passes; one still held at the call trips.
2. **`assertNoSameTxGovTokenInflow`** (corroborating, coarser) — for every protected call, the *gross* LDO transferred to the caller anywhere in the tx must not exceed `maxIntraTxAcquired`. Reads the raw transfer log (`getErc20Transfers`), **not** the reduced per-account deltas — the reduced view nets a borrow+repay flash loan to zero and the actor disappears. Catches power routed through a transiently-funded caller, at the cost of also flagging an actor that legitimately receives the token and acts in the same tx. Arm as defense in depth.

Both fire only via `registerCallTrigger` on the protected selectors, so transactions that don't touch governance pay nothing. The assertion is generic over `(governanceContract, governanceToken, maxIntraTxAcquired, protectedSelectors)`, so it applies to any LDO-balance-weighted governance entrypoint, not just `objectToMotion`.

## Files

- `examples/lido/src/LidoEasyTrackFlashLoanAssertion.sol` — the assertion
- `examples/lido/test/LidoEasyTrackFlashLoanAssertion.t.sol` — 10 pass/trip tests
- `examples/lido/test/LidoGovMocks.sol` — EasyTrack + flash-attacker + honest-objector mocks
- `examples/lido/test/LidoMocks.sol` — gave the shared `MockERC20` standard `Transfer` events (correctness fix so log-based precompiles see transfers; no behavior change to existing suites)

## Testing

`FOUNDRY_PROFILE=lido pcl test` → **69 passed, 0 failed** across all 5 Lido suites (10 new; the other 59 unaffected by the mock edit).

## Note for adoption

EasyTrack isn't upgradable, so to attach this gate Lido registers the EasyTrack address as an assertion adopter on the Credible Layer — no contract change needed.

### DIFF
--- a/examples/lido/src/LidoEasyTrackFlashLoanAssertion.sol
+++ b/examples/lido/src/LidoEasyTrackFlashLoanAssertion.sol
@@ -1,0 +1,163 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.13;
+
+import {PhEvm} from "credible-std/PhEvm.sol";
+import {Assertion} from "credible-std/Assertion.sol";
+import {AssertionSpec} from "credible-std/SpecRecorder.sol";
+
+/// @title LidoEasyTrackFlashLoanAssertion
+/// @author Phylax Systems
+/// @notice Blocks flash-loaned governance power from being exercised against a Lido governance
+///         contract (the motivating case is EasyTrack), without upgrading the governance contract.
+/// @dev Apply to the governance contract itself (e.g. EasyTrack) — the account that receives the
+///      flash-loan-sensitive calls. Configure it with the governance token (LDO) and the set of
+///      function selectors whose effect is weighted by `governanceToken` balance.
+///
+///      ## The vulnerability (from the Lido call)
+///      EasyTrack is optimistic governance: whitelisted factories create *motions* that execute
+///      after an objection window unless LDO holders object past a threshold. An objection's weight
+///      is read from the governance token at the motion's snapshot block, and that snapshot is taken
+///      at the block the motion *starts* rather than the block before. Because a same-block balance
+///      read returns the live balance, an attacker can, in a single transaction, **flash-loan LDO,
+///      exercise the balance-weighted action, and repay the loan** — borrowed voting power that was
+///      never actually held counts. EasyTrack is not upgradable, so the snapshot cannot simply be
+///      moved to `block.number - 1`; an external execution gate is the agile patch.
+///
+///      ## The invariant
+///      Voting power exercised against the protected selectors must be power the actor *already held
+///      at the start of the transaction* — not power acquired within the same transaction. A flash
+///      loan is exactly the act of acquiring (and returning) the token inside one transaction, so an
+///      actor whose `governanceToken` balance is larger at the moment of the governance call than it
+///      was at the start of the transaction is using power it did not durably hold. This re-creates
+///      the "snapshot a block earlier" property from the outside: same-block (flash-loaned) balance
+///      no longer counts, because the transaction that relies on it is never included.
+///
+///      ## Two detection layers (the two approaches discussed on the call), both armable
+///      - **`assertNoFlashLoanedVotingPower` (primary, precise).** For the protected call that fired
+///        this invocation, compare the caller's `governanceToken` balance at the start of the
+///        transaction (`PreTx`) against its balance immediately before the governance call executes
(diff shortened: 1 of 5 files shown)


examples/lido/test/LidoEasyTrackFlashLoanAssertion.t.sol
```diff
@@ -0,0 +1,130 @@
+// SPDX-License-Identifier: MIT
```
mateo-mro, line 1: Can we add the lido profile to the example PCL matrix in CI? The current PR checks only show the fluid example job, so these new EasyTrack tests can regress without a required check catching it.

examples/lido/src/LidoEasyTrackFlashLoanAssertion.sol
```diff
@@ -0,0 +1,160 @@
...
+
+    /// @notice Primary layer: forbids exercising governance power acquired within the transaction.
+    /// @dev For every call to a protected selector, the caller's governance-token balance immediately
+    ///      before the call (`PreCall`) must not exceed its balance at the start of the transaction
+    ///      (`PreTx`) by more than `maxIntraTxAcquired`. Reading at `PreCall` aligns with what the
+    ///      governance snapshot sees: a flash loan repaid before the call reads back to baseline and
+    ///      passes; a flash loan still held at the call shows the inflated balance and trips.
+    function assertNoFlashLoanedVotingPower() external view {
+        PhEvm.ForkId memory preTx = _preTx();
+
```
mateo-mro, line 118: Can we use `ph.context()` here and check only the call that fired this assertion? `registerCallTrigger` already invokes the assertion for each protected call. Walking every selector and every matching call repeats the same balance reads for each protected call in the transaction, and the inflow layer repeats the transfer scan the same way.

examples/lido/test/LidoEasyTrackFlashLoanAssertion.t.sol
makemake-kbo: Done in 87a2936. Added `lido` to the `examples-pcl` matrix in `.github/workflows/solidity-test.yml` (now `[fluid, lido]`), so `FOUNDRY_PROFILE=lido pcl test` runs on every PR. Full lido profile is green locally: 69 tests passed (10 of them the EasyTrack suite).

examples/lido/src/LidoEasyTrackFlashLoanAssertion.sol
makemake-kbo: Done in 87a2936. Both layers now use `registerFnCallTrigger` + `ph.context()` and check only the call that fired the invocation, instead of walking every selector and every matching call. A small `_triggerCaller(ctx)` resolves the firing call's caller by matching `ctx.callStart` against the selector-filtered `getCallInputs` view — I used `getCallInputs` rather than `matchingCalls` so the assertion stays executable under `pcl test`. The primary layer now does one `PreTx`/`PreCall` balance pair per protected call, and the inflow layer scans transfers once for the single firing actor. All 10 EasyTrack tests still pass.

mateo-mro (approved):

PR #83, opened 2026-06-30
feat(safe): add CredibleSafeGuard transaction guard

makemake-kbo:
## Summary

Adds `CredibleSafeGuard`, a Gnosis Safe **transaction guard** that only allows Safe executions while the current block is credible — as reported by the on-chain [Credible Registry](https://github.com/phylaxsystems/credible-registry) — and **fails open** after a configurable window when the credible builder set is offline, so the Safe is never bricked.

Closes ENG-3671.

## How it works

Installed on a Safe via `setGuard`. On every owner-path `execTransaction`, Safe calls `checkTransaction`, which:

1. **Fails open** (allows) if the last credible block lags the current block by more than `failOpenBlockThreshold` — the credible builder set looks offline.
2. Otherwise requires the current block to be credible, reverting `NonCredibleBlock` if not.
3. Allows when the current block is itself credible.

Implements `ITransactionGuard` + ERC-165; `supportsInterface` returns `0xe6d7a83a` (equal to Safe's `type(Guard).interfaceId`), so it passes Safe's `setGuard` `GS300` check.

## Design notes

- **Registry interface**: vendored `ICredibleRegistry` (`isCredibleBlock`, `lastCredibleBlock`) mirroring the registry repo, so the guard is drop-in compatible.
- **"15 minutes" expressed in blocks**: the registry records credibility by block number with no timestamps, so the fail-open window is a configurable block count — the blocks a builder produces in ~15 min on the target chain (e.g. 75 @ 12s, 450 @ 2s). Documented in the Safe README.
- **Configurable**: registry address and fail-open threshold are immutable constructor args; zero values revert.

## Tests

- `test/protection/safe/CredibleSafeGuard.t.sol` — 23 unit tests against a mock registry: allow / block / fail-open, exact boundaries, ERC-165 (incl. EIP-165 `0xffffffff` → false), no-underflow, delegatecall ABI, a `MockSafe` call path, and a fuzz over the decision rule.
- `examples/safe-guard/` — 5 **real Gnosis Safe v1.4.1** end-to-end tests: install via signed `execTransaction` → `setGuard` (exercises Safe's real ERC-165 check), then real signed transactions through the credible / non-credible / **fail-open** / stall→recover paths.

The Safe integration suite runs on its own `safe-guard` profile (legacy pipeline: `optimizer`, no `via_ir`, since Safe does not compile under the root `via_ir` profiles) and is wired into the `examples-pcl` CI matrix. Safe is added as a test-only submodule pinned at `v1.4.1-3`.

```
FOUNDRY_PROFILE=ci forge test --match-path 'test/protection/safe/CredibleSafeGuard.t.sol'   # 23 passed
FOUNDRY_PROFILE=safe-guard forge test                                                        # 5 passed
```

### DIFF
--- a/.github/workflows/solidity-test.yml
+++ b/.github/workflows/solidity-test.yml
@@ -34,12 +34,14 @@ jobs:
   # Example assertion suites live under their own foundry profiles (examples/<name>) and also use the
   # Phorge-only `cl.assertion` cheatcode, so they run under `pcl test` here. Add a profile to the
   # matrix when its example is ready to be gated on every PR.
+  # `safe-guard` is a plain-forge integration suite (real Gnosis Safe + CredibleSafeGuard) on its own
+  # profile because the Safe contracts need the legacy pipeline; pcl runs plain forge tests too.
   examples-pcl:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        profile: [fluid, lido, mellow]
+        profile: [fluid, lido, mellow, safe-guard]
     steps:
       - name: Checkout
         uses: actions/checkout@v6
--- a/src/protection/safe/CredibleSafeGuard.sol
+++ b/src/protection/safe/CredibleSafeGuard.sol
@@ -45,11 +45,19 @@ interface ITransactionGuard is IERC165 {
 
 /// @title CredibleSafeGuard
 /// @author Phylax Systems
-/// @notice Safe transaction guard that only allows Safe executions while the current block is
-///         credible, i.e. built by a Credible Layer builder that enforces assertions. When the
-///         credible builder set is offline the guard fails open so the Safe is never bricked.
+/// @notice Safe transaction guard that only allows owner/multisig Safe transactions while the
+///         current block is credible, i.e. built by a Credible Layer builder that enforces
+///         assertions. When the credible builder set is offline the guard fails open so the Safe
+///         is never bricked.
 /// @dev Installed on a Safe via `setGuard(address(thisGuard))`. The Safe calls
 ///      {checkTransaction} before every owner-path execution; a revert blocks that execution.
+///
+///      Scope. This guard implements only Safe's transaction-guard interface ({ITransactionGuard}),
+///      so it gates the owner/multisig `execTransaction` path only. Module executions
+///      (`execTransactionFromModule`/`...ReturnData`) do not run transaction guards, so an enabled
+///      module can still execute while the current block is not credible. Gating module executions
+///      requires a separate Safe module guard (the v1.5.0 `checkModuleTransaction` hook) or a
+///      Credible Layer assertion such as {SafeTxShapeAssertion}.
 ///
 ///      Decision in {checkTransaction}:
 ///      1. If the credible builder set looks offline (the most recent credible block is more
--- a/src/protection/safe/CredibleSafeGuard.sol
+++ b/src/protection/safe/CredibleSafeGuard.sol
@@ -60,12 +60,13 @@ interface ITransactionGuard is IERC165 {
 ///      Credible Layer assertion such as {SafeTxShapeAssertion}.
 ///
 ///      Decision in {checkTransaction}:
-///      1. If the credible builder set looks offline (the most recent credible block is more
-///         than `failOpenBlockThreshold` blocks behind the current block), FAIL OPEN and allow
-///         the transaction. This prevents a stalled builder set from permanently locking the Safe.
-///      2. Otherwise the builder set is live, so the current block MUST be credible; if it is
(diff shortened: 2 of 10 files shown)


.github/workflows/solidity-test.yml
```diff
@@ -34,12 +34,14 @@ jobs:
...
   # matrix when its example is ready to be gated on every PR.
+  # `safe-guard` is a plain-forge integration suite (real Gnosis Safe + CredibleSafeGuard) on its own
+  # profile because the Safe contracts need the legacy pipeline; pcl runs plain forge tests too.
   examples-pcl:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        profile: [fluid]
+        profile: [fluid, safe-guard]
```
mateo-mro: This drops the existing `lido` example from the CI matrix while adding the new Safe guard suite. `examples/lido` still exists, and the PR does not explain removing that coverage, so this looks like an unrelated validation regression. Please add the new Safe guard profile without removing `lido`.

src/protection/safe/CredibleSafeGuard.sol
```diff
@@ -0,0 +1,151 @@
...
+        address msgSender
+    ) external;
+
+    /// @notice Called by the Safe after executing the transaction.
+    function checkAfterExecution(bytes32 hash, bool success) external;
+}
+
+/// @title CredibleSafeGuard
+/// @author Phylax Systems
+/// @notice Safe transaction guard that only allows Safe executions while the current block is
```
mateo-mro: `checkTransaction` only covers the Safe transaction guard path. Module executions go through the module execution path, so an enabled module can still execute while the current block is not credible unless another module guard or assertion covers it. Since ENG 3671 says Safe transactions, I think this should either cover that path too or make the owner path scope explicit wherever the guard is described.

makemake-kbo: intended

.github/workflows/solidity-test.yml
makemake-kbo: This was the diff against an older base. The original head (8f92dfa) branched from when the matrix was just `[fluid]` and changed it to `[fluid, safe-guard]`, so GitHub rendered it against current master (which had since gained `lido` and `mellow`) as if those were being dropped. The branch is now rebased onto current master, and the matrix is `[fluid, lido, mellow, safe-guard]` — `lido` and `mellow` are retained and `safe-guard` is added alongside them. Nothing is dropped.

src/protection/safe/CredibleSafeGuard.sol
makemake-kbo: Made the owner-path scope explicit in 8499126, per your second option. The contract NatSpec now has a `Scope` note stating it implements only `ITransactionGuard` and gates the owner/multisig `execTransaction` path, that module executions (`execTransactionFromModule`/`...ReturnData`) bypass transaction guards and can still execute in a non-credible block, and that gating modules needs a separate module guard (`checkModuleTransaction`) or an assertion like `SafeTxShapeAssertion`. The README's guard section gets a matching `### Scope` subsection and the one-line summary now says owner-path.

src/protection/safe/CredibleSafeGuard.sol
```diff
@@ -0,0 +1,159 @@
...
+
+    /// @notice Whether the guard is currently failing open because the builder set looks offline.
+    /// @return True if the most recent credible block lags the current block beyond the threshold.
+    function failOpenActive() external view returns (bool) {
+        return _failOpenActive();
+    }
+
+    /// @dev Core gate: fail open when the builder set is offline, otherwise require credibility.
+    function _checkCredibleBlock() internal view {
+        if (_failOpenActive()) return;
```
mateo-mro: On the normal allowed path where the current block is credible, this reads the registry twice: first `lastCredibleBlock()` through `_failOpenActive()`, then `isCredibleBlock(block.number)`. Checking `isCredibleBlock(block.number)` first and only reading `lastCredibleBlock()` when the current block is not credible would keep the same allow logic while saving one external static call on the expected hot path.

makemake-kbo: Done in 1ad7e97 — `_checkCredibleBlock` now checks `isCredibleBlock(block.number)` first and returns early, only calling `lastCredibleBlock()` via `_failOpenActive()` when the current block is not credible, so the expected hot path costs a single registry call. Applied the same reordering to the `isCurrentBlockAllowed` view helper for consistency, and updated the NatSpec decision list plus the README to match the evaluation order. All guard unit tests and the safe-guard integration profile pass.

mateo-mro (approved):

PR #84, opened 2026-06-30
feat(credible-std): add reusable CredibleBlockGuard onlyCredibleBlock modifier

makemake-kbo:
## Summary

Adds **`CredibleBlockGuard`** — a reusable mixin that gates functions on **block credibility**. Inherit it and apply the `onlyCredibleBlock` modifier to any function that should only execute while the current block is credible (built by a Credible Layer builder that enforces assertions), with a **fail-open** so protected contracts are never permanently bricked.

This is the general-purpose, function-level form of the credibility gate (ENG-3663). `CredibleSafeGuard` (ENG-3671, #83) is the same decision wired into a Gnosis Safe transaction guard; this one is just a modifier with no Safe coupling.

```solidity
contract MyVault is CredibleBlockGuard {
    constructor(ICredibleRegistry reg, uint256 failOpenThreshold)
        CredibleBlockGuard(reg, failOpenThreshold) {}

    function withdraw(uint256 amt) external onlyCredibleBlock { ... }
}
```

## Decision (`onlyCredibleBlock`)

1. **Fail open** — if the most recent credible block lags the current block by more than `failOpenBlockThreshold`, the builder set looks offline → allow the call.
2. Otherwise the builder set is live → the current block **must** be credible, else revert `NonCredibleBlock`.
3. A credible current block is always allowed.

`isCurrentBlockAllowed()` / `failOpenActive()` expose the same decision as view helpers.

## ABI compatibility

`ICredibleRegistry` here is identical to the deployed [`phylaxsystems/credible-registry`](https://github.com/phylaxsystems/credible-registry) interface (`isCredibleBlock(uint256)`, `lastCredibleBlock()`) — verified against the source, same selectors. The registry address and fail-open threshold are configured per deployment via the constructor; the threshold is expressed as a block count approximating the chain's ~15-minute budget (e.g. ~75 blocks at 12s).

## Files

- `src/protection/credible_block/CredibleBlockGuard.sol` — the mixin / modifier
- `src/protection/credible_block/ICredibleRegistry.sol` — registry read interface
- `src/protection/credible_block/README.md` — usage + fail-open window docs
- `test/protection/credible_block/CredibleBlockGuard.t.sol` — unit + fuzz suite

## Testing

```
forge test --match-path "test/protection/credible_block/CredibleBlockGuard.t.sol"
15 passed; 0 failed; 0 skipped
```

Covers construction guards, the credible/non-credible/fail-open boundaries, builder stall-then-recover, underflow defense, modifier scope (guarded vs unguarded functions), and a spec-matching fuzz test.

## Note for reviewers

The upstream `credible-registry` repo also ships a `CredibleBlockGuard` mixin. This implementation matches its core fail-open math exactly but differs in API: the threshold is constructor-injected (to satisfy the per-chain 15-minute requirement) rather than a `virtual` override, and there's no per-sender bypass allowlist. Happy to align with upstream's override/bypass API if preferred. There is also a near-identical `ICredibleRegistry` introduced by #83 under `src/protection/safe/`; on merge these should be consolidated (e.g. `CredibleSafeGuard is CredibleBlockGuard`).

Closes ENG-3663.

### DIFF
--- a/test/protection/credible_block/CredibleBlockGuard.t.sol
+++ b/test/protection/credible_block/CredibleBlockGuard.t.sol
@@ -216,28 +216,48 @@ contract CredibleBlockGuardTest is Test {
     }
 
     // ---------------------------------------------------------------------
-    // Fuzz: decision is exactly "credible current block OR gap beyond threshold"
+    // Decision table: allowed == "credible current block OR gap beyond threshold"
+    //
+    // The gate is fully deterministic, so a fixed table over the boundary-relevant
+    // gaps (below / at / above the threshold, with the current block credible or
+    // not) gives the same coverage an equivalent fuzz test would. It is kept
+    // table-driven rather than fuzzed because this directory is gated by the
+    // `pcl test` runner, whose fuzzing worker overflows its stack and aborts the
+    // whole protection job.
     // ---------------------------------------------------------------------
 
-    function testFuzz_decisionMatchesSpec(uint256 gap, bool currentCredible) public {
-        gap = bound(gap, 0, 10 * THRESHOLD);
-        // Keep BASE_BLOCK large enough that block.number - gap never underflows.
-        registry.setLastCredibleBlock(block.number - gap);
-        if (currentCredible) registry.setCredibleBlock(block.number, true);
+    function test_decisionMatchesSpec_acrossGaps() public {
+        uint256[6] memory gaps = [uint256(0), 1, THRESHOLD - 1, THRESHOLD, THRESHOLD + 1, 10 * THRESHOLD];
+
+        for (uint256 i = 0; i < gaps.length; i++) {
+            _assertDecision(gaps[i], false);
+            _assertDecision(gaps[i], true);
+        }
+    }
+
+    /// @dev Asserts the gate's decision for one (gap, currentCredible) case. Uses fresh
+    ///      instances so each case starts from a clean slate, exactly as a fuzz run would.
+    function _assertDecision(uint256 gap, bool currentCredible) internal {
+        MockCredibleRegistry reg = new MockCredibleRegistry();
+        GuardedVault v = new GuardedVault(reg, THRESHOLD);
+
+        // BASE_BLOCK is large enough that block.number - gap never underflows.
+        reg.setLastCredibleBlock(block.number - gap);
+        if (currentCredible) reg.setCredibleBlock(block.number, true);
 
         bool failOpen = gap > THRESHOLD; // block.number > last is guaranteed for gap > 0
         bool expectedAllowed = failOpen || currentCredible;
 
-        assertEq(vault.failOpenActive(), failOpen);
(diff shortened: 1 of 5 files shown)


test/protection/credible_block/CredibleBlockGuard.t.sol
```diff
@@ -0,0 +1,243 @@
...
+        vault.doProtectedAction();
+
+        assertEq(vault.actions(), 3);
+    }
+
+    // ---------------------------------------------------------------------
+    // Fuzz: decision is exactly "credible current block OR gap beyond threshold"
+    // ---------------------------------------------------------------------
+
+    function testFuzz_decisionMatchesSpec(uint256 gap, bool currentCredible) public {
```
mateo-mro: This new suite is now part of the protection job, and that job is red on this head. The log shows `fatal runtime error: stack overflow, aborting` and exits 134 during the protection test step. Please get this check green before merge so the guard is covered by the same gate as the rest of this directory.

makemake-kbo: Fixed in 162082f. Root cause was the pcl 1.5.0 test runner that gates this directory: its fuzzing worker overflows its stack and aborts the whole `protection-pcl` job (the `fatal runtime error: stack overflow` you saw is a Rust-level abort in a worker thread, interleaved with other suites' output — not an EVM/assertion failure). This suite added the only fuzz test under `test/protection/**`, which is why the gate went red exactly on this head.

The gate's decision is fully deterministic (`allowed == credible-now OR gap > threshold`), so I replaced the fuzz test with a table-driven `test_decisionMatchesSpec_acrossGaps` that enumerates the boundary-relevant gaps (0, 1, threshold-1, threshold, threshold+1, 10*threshold) with the current block both credible and not. Same coverage, no fuzzer. The suite stays in the protection gate as requested; 15/15 pass under `pcl test`.

test/protection/credible_block/CredibleBlockGuard.t.sol
```diff
@@ -0,0 +1,264 @@
...
+    // The gate is fully deterministic, so a fixed table over the boundary-relevant
+    // gaps (below / at / above the threshold, with the current block credible or
+    // not) gives the same coverage an equivalent fuzz test would. It is kept
+    // table-driven rather than fuzzed because this directory is gated by the
+    // `pcl test` runner, whose fuzzing worker overflows its stack and aborts the
+    // whole protection job.
+    // ---------------------------------------------------------------------
+
+    function test_decisionMatchesSpec_acrossGaps() public {
+        uint256[6] memory gaps =
```
mateo-mro: The Forge formatter check is failing on this table declaration. CI wants the array declaration on one line, so please run the formatter before merge.

mateo-mro (dismissed):

mateo-mro (approved):

PR #85, opened 2026-07-02
test: arm matchingCalls through cl.assertion instead of routing around it

makemake-kbo:
## Why

`ph.matchingCalls` shipped in the PhEvm interface, the docs, and multiple protection suites in April — and no executor implemented it until phylaxsystems/credible-sdk#1137. Nothing in CI noticed for months because every test that touches it routes around the precompile: the fluid suite's own header comment said the query "is not simulated by local `pcl test`" and validated the decode policy through a harness fed hand-built bytes, and the perpetual/lending suites were switched to `getAllCallInputs`. The first thing that ever executed `matchingCalls` was a user's assertion, which reverted with `Precompile selector not found: 0x0dc1cc5d`.

This PR makes the tests execute the precompile for real, so interface/executor drift fails CI instead of surfacing as a user report.

## Changes

- **`test/protection/MatchingCallsPrecompile.t.sol`** (new): armed battery via `cl.assertion` pinning the full `TriggerCall` contract — depth/parent chaining across nested calls, execution order, selector-stripped `input`, `topLevelOnly`/`minDepth`/`limit` filters, reverted-subtree invisibility, and the `Assertion._matchingCalls` helper path.
- **`examples/fluid`**: new `FluidFlowBreakerArmedTest` arms the breaker's real `matchingCalls` scan + borrow policy through a call-trigger subclass (`ArmedBreakerAssertion`). Everything downstream of `ph.outflowContext()` — filter, precompile query, selector-stripped decode — is the production path against real recorded calldata; only the outflow trigger context stays harness-side (it's executor-driven and can't be armed locally). The borrow-blocked test expects the policy's **exact** revert message, so a missing precompile can't pass for the wrong reason. The misleading header comment is rewritten, and `triggers()` on the breaker is made `virtual` for the test subclass.
- Pure-policy harness tests are kept — they still cover decode edge cases cheaply.

## ⚠️ CI will be red until pcl ships the precompile — by design

The `protection-pcl` and `examples-pcl` (fluid) jobs pin pcl 1.5.0, which lacks the precompile, so the new tests fail there with `Precompile selector not found: 0x0dc1cc5d`. That failure is exactly the signal this PR exists to produce. Merge order: land phylaxsystems/credible-sdk#1137 → phylaxsystems/phoundry#88 → phylaxsystems/pcl#106, cut a pcl release, bump the pcl version in `.github/workflows`, then merge this.

## Verification

Against a pcl built from the pin-bump branch (pcl#106):
- `FOUNDRY_PROFILE=ci pcl test --match-path 'test/protection/**'` — 88/88 pass (5 new).
- `FOUNDRY_PROFILE=fluid pcl test` — 26/26 pass (4 new armed).

Against released pcl (no precompile): every armed test fails with `Precompile selector not found: 0x0dc1cc5d`, including the expectRevert test (message mismatch).

### DIFF
--- a/test/protection/MatchingCallsPrecompile.t.sol
+++ b/test/protection/MatchingCallsPrecompile.t.sol
@@ -1,0 +1,200 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.13;
+
+import {Test} from "forge-std/Test.sol";
+
+import {CredibleTest} from "../../src/CredibleTest.sol";
+import {Assertion} from "../../src/Assertion.sol";
+import {PhEvm} from "../../src/PhEvm.sol";
+import {AssertionSpec} from "../../src/SpecRecorder.sol";
+
+/// @notice Armed end-to-end battery for the `ph.matchingCalls` precompile.
+/// @dev Every test here executes `ph.matchingCalls` through a real `cl.assertion` arming — never a
+///      harness fed with hand-built bytes — so a pcl/executor build without the precompile fails
+///      loudly with `Precompile selector not found: 0x0dc1cc5d` instead of staying silently green.
+///      This is the regression test for the gap where the cheatcode shipped in the interface, docs,
+///      and protection suites for months while no executor implemented it.
+contract MatchingCallsTarget {
+    uint256 public value;
+
+    /// @dev Fans out two nested self-calls so the trace has same-selector calls at depth > 1,
+    ///      and a caught reverted call whose subtree must be invisible to the precompile.
+    function driver() external {
+        try this.pokeReverting() {
+            revert("expected pokeReverting to revert");
+        } catch {}
+
+        this.outer(7);
+    }
+
+    function outer(uint256 seed) external {
+        this.inner(seed + 1);
+        this.inner(seed + 2);
+    }
+
+    function inner(uint256 v) external {
+        value = v;
+    }
+
+    function pokeReverting() external {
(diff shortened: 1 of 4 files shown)


test/protection/MatchingCallsPrecompile.t.sol
```diff
@@ -0,0 +1,200 @@
...
+            ph.matchingCalls(adopter, MatchingCallsTarget.driver.selector, _anyFilter(), 32);
+
+        require(calls.length == 1, "driver: expected exactly one call");
+        require(calls[0].target == adopter, "driver: target mismatch");
+        require(calls[0].selector == MatchingCallsTarget.driver.selector, "driver: selector mismatch");
+        require(calls[0].depth == 1, "driver: depth != 1");
+        require(calls[0].parentCallId == NO_PARENT, "driver: top-level call must have no parent");
+        require(calls[0].callType == 1, "driver: callType != CALL");
+        require(calls[0].success, "driver: success != true");
+        require(calls[0].input.length == 0, "driver: input must be empty (no args)");
```
mateo-mro, line 79: Since this file is meant to pin the `TriggerCall` shape, I would also assert `caller`, `codeAddress`, and `value` here. Right now a decoder or executor bug in those fields would still keep this battery green, even though the PR description calls out the full `TriggerCall` contract.

mateo-mro (dismissed):

mateo-mro (approved):

PR #86, opened 2026-07-08
feat(balancer): Balancer V3 vault assertion suite

makemake-kbo:
## Summary

Example assertion bundle for Balancer V3's singleton Vault under `examples/balancer/`, following the per-protocol profile layout (`FOUNDRY_PROFILE=balancer`).

Balancer V3 concentrates every pool's custody, balances, and BPT supply in one Vault and trusts three things it never re-checks: the pool's own swap math (`onSwap` output is applied with no invariant guard), the mutual consistency of its layered ledgers (`_reservesOf` vs pool balances vs aggregate fees), and rate-provider answers for `WITH_RATE` tokens. The suite asserts those externally:

- **`BalancerV3VaultAssertion`** (one watched pool inside the singleton, PoolKey-style calldata filtering like the UniswapV4 example):
  - `assertSwapPreservesPoolInvariant` — call-scoped on `Vault.swap`: the pool's own `computeInvariant` over live balances may not decrease (named dust tolerance), BPT supply is frozen, balances move in the swap's direction.
  - `assertVaultCustodyCoversPoolAccounting` — tx-end: `token.balanceOf(vault) >= getReservesOf(token) >= pool raw balance + aggregate swap/yield fees`.
  - `assertTokenRatesWithinDriftBound` — tx-end: rate-provider `getRate()` stays nonzero and within a bps bound across the transaction.
- **`BalancerV3VaultOutflowAssertion`** — rolling-window `watchCumulativeOutflow` breaker on a token leaving Vault custody (flow-based drain protection, mellow pattern).

All thresholds (invariant dust, rate drift, outflow cap/window) are constructor parameters for per-deployment calibration.

## Notes

- Swap-leg decoding reads `pool`/`tokenIn`/`tokenOut` at fixed calldata offsets (with a bound-checked head offset) instead of abi-decoding the full `VaultSwapParams`: the full decode plus byte-copy costs enough gas to blow the assertion budget on the honest path.
- Verified against `balancer-v3-monorepo` sources; getters are reachable on the Vault address via its extension fallback.

## Testing

`FOUNDRY_PROFILE=balancer pcl test` — 20/20 passing: one honest-pass and one tripping test per invariant, each driven by a single mock failure knob, plus breaker direct-invocation and constructor-guard tests.

### DIFF
--- a/examples/balancer/src/BalancerV3VaultAssertion.sol
+++ b/examples/balancer/src/BalancerV3VaultAssertion.sol
@@ -55,8 +55,9 @@ contract BalancerV3VaultAssertion is BalancerV3VaultHelpers {
         registerTxEndTrigger(this.assertTokenRatesWithinDriftBound.selector);
     }
 
-    /// @notice A swap on a hookless pool must grow (or at worst preserve) the pool invariant and
-    ///         move its live tokenOut balance in the swap's direction.
+    /// @notice A swap on a hookless pool must grow (or at worst preserve) the pool invariant,
+    ///         leave BPT supply unchanged, and move its live tokenOut balance in the swap's
+    ///         direction.
     /// @dev The Vault takes `onSwap`'s output on trust: no `require` recomputes the curve. This
     ///      check recomputes the pool's invariant from live balances through the pool's own math at
     ///      the pre-call and post-call forks. With any nonzero swap fee the invariant must grow, so
@@ -107,6 +108,8 @@ contract BalancerV3VaultAssertion is BalancerV3VaultHelpers {
         uint256 preInvariant = _invariantOf(preLiveBalances, pre);
         uint256 postInvariant = _invariantOf(postLiveBalances, post);
         require(postInvariant + INVARIANT_DUST_TOLERANCE >= preInvariant, "BalancerV3: swap decreased pool invariant");
+
+        require(_bptTotalSupplyAt(post) == _bptTotalSupplyAt(pre), "BalancerV3: swap changed BPT supply");
 
         uint256 indexOut = _tokenIndex(preSnap.tokens, tokenOut);
         require(
--- a/examples/balancer/src/BalancerV3VaultAssertion.sol
+++ b/examples/balancer/src/BalancerV3VaultAssertion.sol
@@ -46,11 +46,23 @@ contract BalancerV3VaultAssertion is BalancerV3VaultHelpers {
 
     /// @notice Registers Vault selectors against their protection assertions.
     /// @dev The Vault is the assertion adopter. The swap check is call-scoped so it compares the
-    ///      exact pre-call and post-call snapshots of the matched swap; the custody and rate checks
-    ///      are transaction-wide envelopes because add/remove liquidity, buffer operations, and fee
-    ///      collection all move the same accounting.
+    ///      exact pre-call and post-call snapshots of the matched operation. The transaction-end
+    ///      checks first inspect calldata and return before reading snapshots unless swap or
+    ///      liquidity traffic touched the watched pool.
     function triggers() external view override {
         registerFnCallTrigger(this.assertSwapPreservesPoolInvariant.selector, IBalancerV3VaultLike.swap.selector);
+        registerFnCallTrigger(
+            this.assertOperationRatesWithinBaseline.selector, IBalancerV3VaultLike.addLiquidity.selector
+        );
+        registerFnCallTrigger(
+            this.assertOperationRatesWithinBaseline.selector, IBalancerV3VaultLike.removeLiquidity.selector
+        );
+        registerFnCallTrigger(
+            this.assertOperationRatesWithinBaseline.selector, IBalancerV3VaultLike.initialize.selector
+        );
+        registerFnCallTrigger(
+            this.assertOperationRatesWithinBaseline.selector, IBalancerV3VaultLike.disableRecoveryMode.selector
+        );
         registerTxEndTrigger(this.assertPoolAccountingWithinVaultCustody.selector);
         registerTxEndTrigger(this.assertTokenRatesWithinDriftBound.selector);
     }
@@ -125,6 +137,35 @@ contract BalancerV3VaultAssertion is BalancerV3VaultHelpers {
     ///      compare; a provider that answered zero is a broken baseline and fails instead of being
     ///      conflated with the deployment case.
     function _requireSwapRatesWithinBaseline(PoolTokenSnapshot memory snap, PhEvm.ForkId memory pre) internal view {
+        _requireRatesWithinBaseline(snap, pre, "BalancerV3: swap priced against rate beyond drift bound");
+    }
+
+    /// @notice Pins rates consumed by liquidity and recovery-resync operations to their transaction baseline.
+    /// @dev This closes the router-shaped manipulate-operation-restore gap for both liquidity
+    ///      selectors. A pool with state-changing liquidity hooks needs a pool-specific assertion
+    ///      that observes the rate after its before hook, since an outer call snapshot cannot see a
+    ///      rate moved and restored entirely inside that hook lifecycle.
+    function assertOperationRatesWithinBaseline() external view {
+        PhEvm.TriggerContext memory ctx = ph.context();
+        bytes memory input = ph.callinputAt(ctx.callStart);
+        address pool = ctx.selector == IBalancerV3VaultLike.addLiquidity.selector
+            || ctx.selector == IBalancerV3VaultLike.removeLiquidity.selector
+            ? _operationPool(input)
+            : _firstAddressArg(input);
+        if (pool != POOL) {
+            return;
+        }
+
+        PhEvm.ForkId memory pre = _preCall(ctx.callStart);
+        PoolTokenSnapshot memory preSnap = _poolTokenSnapshotAt(pre);
+        _requireRatesWithinBaseline(preSnap, pre, "BalancerV3: liquidity priced against rate beyond drift bound");
+    }
+
+    function _requireRatesWithinBaseline(
+        PoolTokenSnapshot memory snap,
+        PhEvm.ForkId memory pre,
+        string memory driftError
+    ) internal view {
         PhEvm.ForkId memory preTx = _preTx();
         for (uint256 i; i < snap.tokenInfo.length; ++i) {
             address rateProvider = snap.tokenInfo[i].rateProvider;
@@ -132,17 +173,14 @@ contract BalancerV3VaultAssertion is BalancerV3VaultHelpers {
                 continue;
             }
 
-            uint256 rateAtSwap = _rateAt(rateProvider, pre);
+            uint256 rateAtOperation = _rateAt(rateProvider, pre);
 
             (bool answered, uint256 baseline) = _rateStatusAt(rateProvider, preTx);
             if (!answered) {
                 continue;
             }
             require(baseline != 0, "BalancerV3: zero rate baseline");
-            require(
-                _absDiff(rateAtSwap, baseline) <= _bpsOf(baseline, RATE_DRIFT_TOLERANCE_BPS),
-                "BalancerV3: swap priced against rate beyond drift bound"
-            );
+            require(_absDiff(rateAtOperation, baseline) <= _bpsOf(baseline, RATE_DRIFT_TOLERANCE_BPS), driftError);
         }
     }
 
@@ -160,6 +198,9 @@ contract BalancerV3VaultAssertion is BalancerV3VaultHelpers {
     ///      accounting inflated beyond anything custody could pay out).
     function assertPoolAccountingWithinVaultCustody() external view {
         _requireConfiguredVaultIsAdopter();
+        if (!_transactionTouchesWatchedPool()) {
+            return;
+        }
         PhEvm.ForkId memory post = _postTx();
         if (!_isPoolInitializedAt(post)) {
             return;
@@ -205,6 +246,9 @@ contract BalancerV3VaultAssertion is BalancerV3VaultHelpers {
     ///        registration (pool registered within this transaction) skips the drift comparison.
     function assertTokenRatesWithinDriftBound() external view {
         _requireConfiguredVaultIsAdopter();
+        if (!_transactionTouchesWatchedPool()) {
+            return;
+        }
         PhEvm.ForkId memory preTx = _preTx();
         PhEvm.ForkId memory postTx = _postTx();
         if (!_isPoolInitializedAt(postTx)) {
@@ -272,6 +316,39 @@ contract BalancerV3VaultAssertion is BalancerV3VaultHelpers {
         }
         for (uint256 i; i < postSnap.tokens.length; ++i) {
             if (_aggregateFeesAt(postSnap.tokens[i], preTx) != _aggregateFeesAt(postSnap.tokens[i], postTx)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /// @dev Cheap tx-end gate. `getCallInputs` reads the trace without any Vault snapshots, so
+    ///      unrelated singleton traffic returns before the per-token accounting and provider work.
+    function _transactionTouchesWatchedPool() internal view returns (bool) {
+        return _structSelectorTouchesWatchedPool(IBalancerV3VaultLike.swap.selector, 32)
+            || _structSelectorTouchesWatchedPool(IBalancerV3VaultLike.addLiquidity.selector, 0)
+            || _structSelectorTouchesWatchedPool(IBalancerV3VaultLike.removeLiquidity.selector, 0)
+            || _addressSelectorTouchesWatchedPool(IBalancerV3VaultLike.initialize.selector)
+            || _addressSelectorTouchesWatchedPool(IBalancerV3VaultLike.removeLiquidityRecovery.selector)
+            || _addressSelectorTouchesWatchedPool(IBalancerV3VaultLike.collectAggregateFees.selector)
+            || _addressSelectorTouchesWatchedPool(IBalancerV3VaultLike.disableRecoveryMode.selector);
+    }
+
+    function _structSelectorTouchesWatchedPool(bytes4 selector, uint256 poolFieldOffset) internal view returns (bool) {
+        PhEvm.CallInputs[] memory calls = ph.getCallInputs(VAULT, selector);
+        for (uint256 i; i < calls.length; ++i) {
+            if (_operationPoolFromArgs(calls[i].input, poolFieldOffset) == POOL) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    function _addressSelectorTouchesWatchedPool(bytes4 selector) internal view returns (bool) {
+        PhEvm.CallInputs[] memory calls = ph.getCallInputs(VAULT, selector);
+        for (uint256 i; i < calls.length; ++i) {
+            (address pool) = abi.decode(calls[i].input, (address));
+            if (pool == POOL) {
                 return true;
             }
         }
--- a/examples/balancer/src/BalancerV3VaultAssertion.sol
+++ b/examples/balancer/src/BalancerV3VaultAssertion.sol
@@ -28,9 +28,19 @@ import {IBalancerV3VaultLike} from "./BalancerV3VaultInterfaces.sol";
 ///      watched pool and silently ignores swaps in other pools. Tx-end checks read only the watched
 ///      pool's accounting, so unrelated pool activity cannot produce false positives.
 contract BalancerV3VaultAssertion is BalancerV3VaultHelpers {
-    constructor(address vault_, address pool_, uint256 invariantDustTolerance_, uint256 rateDriftToleranceBps_)
-        BalancerV3VaultHelpers(vault_, pool_, invariantDustTolerance_, rateDriftToleranceBps_)
-    {
+    /// @notice Whether the watched pool was registered with a before- or after-swap hook.
+    /// @dev Balancer fixes hook wiring when a pool is registered, so classifying it once at
+    ///      deployment avoids an expensive full hook-config read on every swap.
+    bool internal immutable POOL_HAS_SWAP_HOOKS;
+
+    constructor(
+        address vault_,
+        address pool_,
+        bool poolHasSwapHooks_,
+        uint256 invariantDustTolerance_,
+        uint256 rateDriftToleranceBps_
+    ) BalancerV3VaultHelpers(vault_, pool_, invariantDustTolerance_, rateDriftToleranceBps_) {
+        POOL_HAS_SWAP_HOOKS = poolHasSwapHooks_;
         registerAssertionSpec(AssertionSpec.Reshiram);
     }
 
@@ -46,24 +56,24 @@ contract BalancerV3VaultAssertion is BalancerV3VaultHelpers {
     }
 
     /// @notice A swap on a hookless pool must grow (or at worst preserve) the pool invariant and
-    ///         touch nothing else.
+    ///         move its live tokenOut balance in the swap's direction.
     /// @dev The Vault takes `onSwap`'s output on trust: no `require` recomputes the curve. This
     ///      check recomputes the pool's invariant from live balances through the pool's own math at
     ///      the pre-call and post-call forks. With any nonzero swap fee the invariant must grow, so
-    ///      it may never drop by more than the configured absolute rounding dust. BPT supply must
-    ///      be frozen across a swap, and raw pool balances must move in the swap's direction
-    ///      (tokenIn in, tokenOut out).
+    ///      it may never drop by more than the configured absolute rounding dust. The
+    ///      fee-adjusted live tokenOut balance may not increase. Together with invariant
+    ///      non-decrease, this implies nonnegative value entered on the tokenIn side for
+    ///      Balancer's monotonic pool invariants without a redundant input-leg comparison.
     ///
     ///      Scope and trust boundaries, stated exactly:
-    ///      - Pools with before/after-swap hooks are SKIPPED. Those hooks run inside the swap call
-    ///        scope, outside the Vault's internal reentrancy guard, and may legitimately reenter
-    ///        (nested liquidity, donations, nested swaps), so call-boundary snapshots cannot
-    ///        isolate the core swap. Hooked pools need a variant built for their specific hook.
-    ///      - For the supported hookless pools no user code can run inside the swap call, so every
-    ///        registered rate must be identical at the call's boundary snapshots; that is enforced
-    ///        below, which also pins one fixed rate vector under both invariant evaluations and
-    ///        makes raw-balance direction equivalent to live-balance direction. Each rate observed
-    ///        at swap start — the rate the Vault actually prices against — must additionally sit
+    ///      - Pools configured at assertion deployment as having before/after-swap hooks are
+    ///        SKIPPED. Those hooks run inside the swap call scope, outside the Vault's internal
+    ///        reentrancy guard, and may legitimately reenter (nested liquidity, donations, nested
+    ///        swaps), so call-boundary snapshots cannot isolate the core swap. Hooked pools need a
+    ///        variant built for their specific hook.
+    ///      - Supported hookless pools expose no state-changing callback inside the swap, so the
+    ///        registered rate vector cannot move between its call-boundary snapshots. Each rate
+    ///        observed at swap start — the rate the Vault actually prices against — must sit
     ///        within the drift bound of its pre-transaction baseline, so a rate transiently moved
     ///        between calls of one transaction and restored before tx-end is still caught at the
     ///        exact operation that consumed it.
@@ -75,9 +85,12 @@ contract BalancerV3VaultAssertion is BalancerV3VaultHelpers {
     ///        pinning trusted factory/codehash combinations and re-implementing the invariant math
     ///        independently per pool type, which is out of scope for this example.
     function assertSwapPreservesPoolInvariant() external view {
+        if (POOL_HAS_SWAP_HOOKS) {
+            return;
+        }
+
         PhEvm.TriggerContext memory ctx = ph.context();
-        _requireConfiguredVaultIsAdopter();
-        (address pool, address tokenIn, address tokenOut) = _swapArgs(ph.callinputAt(ctx.callStart));
+        (address pool, address tokenOut) = _swapPoolAndTokenOut(ph.callinputAt(ctx.callStart));
         if (pool != POOL) {
             return;
         }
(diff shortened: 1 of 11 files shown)


examples/balancer/src/BalancerV3VaultAssertion.sol
```diff
@@ -0,0 +1,146 @@
...
+        for (uint256 i; i < snap.tokenInfo.length; ++i) {
+            address rateProvider = snap.tokenInfo[i].rateProvider;
+            if (rateProvider == address(0)) {
+                continue;
+            }
+
+            uint256 postRate = _rateAt(rateProvider, postTx);
+            require(postRate != 0, "BalancerV3: rate provider returned zero rate");
+
+            uint256 preRate = _rateAt(rateProvider, preTx);
```
mateo-mro: This should use `_rateOrZeroAt` for the pre transaction read. As written, a pool whose rate provider only exists after the transaction will still revert inside `_rateAt` before the `preRate == 0` branch can run, so the documented registration case is not actually tolerated. A small test with a provider deployed during the transaction would catch this.

examples/balancer/src/BalancerV3VaultAssertion.sol
```diff
@@ -0,0 +1,146 @@
...
+        uint256[] memory preLive = _liveBalancesAt(pre);
+        uint256[] memory postLive = _liveBalancesAt(post);
+        uint256 preInvariant = _invariantOf(preLive, pre);
+        uint256 postInvariant = _invariantOf(postLive, post);
+        require(
+            postInvariant + _bpsOf(preInvariant, INVARIANT_DUST_TOLERANCE_BPS) >= preInvariant,
+            "BalancerV3: swap decreased pool invariant"
+        );
+
+        require(_bptTotalSupplyAt(post) == _bptTotalSupplyAt(pre), "BalancerV3: swap changed BPT supply");
```
fredo: Balancer swap hooks are intentionally reentrant: `onBeforeSwap` / `onAfterSwap` execute outside the internal `_swap` reentrancy guard, and the Vault explicitly reloads balances and rates after the before hook because the hook may change them. A valid hook can therefore perform nested liquidity on this pool (changing BPT supply), or a nested swap/donation (changing the net token directions). Since these snapshots cover the whole outer `Vault.swap` call, this check and the direction checks below can reject valid Balancer executions. Please either restrict and validate the supported hook configuration, or isolate the core swap deltas; `callStart` / `callEnd` do not isolate the exact swap operation. See Balancer Vault.sol lines 215-240 and 269-283: https://github.com/balancer/balancer-v3-monorepo/blob/7861ea2785b96dd10681ff1b8dfe56b36cc202b6/pkg/vault/contracts/Vault.sol#L215-L240

examples/balancer/src/BalancerV3VaultAssertion.sol
```diff
@@ -0,0 +1,146 @@
...
+        PoolTokenSnapshot memory snap = _poolTokenSnapshotAt(post);
+        for (uint256 i; i < snap.tokens.length; ++i) {
+            address token = snap.tokens[i];
+            uint256 reserves = _reservesOfAt(token, post);
+
+            require(
+                _readBalanceAt(token, VAULT, post) >= reserves, "BalancerV3: vault reserves exceed real token custody"
+            );
+            require(
+                reserves >= snap.balancesRaw[i] + _aggregateFeesAt(token, post),
```
fredo, line 218: `reserves` is the global token reserve for the entire Vault singleton, but the right-hand side includes only the watched pool. This therefore cannot establish that reserves cover aggregate pool accounting: other pools and ERC-4626 buffers can consume the same reserve headroom. For example, custody/reserves = 1,000, watched-pool liabilities = 900, and other pool/buffer liabilities = 500; this check passes even though total liabilities are 1,400. Please either aggregate every pool and buffer claim for each token, replace this with a pool-local conservation check, or narrow the name/docs to describe it as only a weak necessary per-pool bound. The current single-pool mock does not exercise this shared-headroom case. Balancer stores pool balances, global reserves, and buffer balances separately here: https://github.com/balancer/balancer-v3-monorepo/blob/7861ea2785b96dd10681ff1b8dfe56b36cc202b6/pkg/vault/contracts/VaultStorage.sol#L88-L152

examples/balancer/src/BalancerV3VaultAssertion.sol
```diff
@@ -0,0 +1,146 @@
...
+        if (pool != POOL) {
+            return;
+        }
+
+        PhEvm.ForkId memory pre = _preCall(ctx.callStart);
+        PhEvm.ForkId memory post = _postCall(ctx.callEnd);
+
+        uint256[] memory preLive = _liveBalancesAt(pre);
+        uint256[] memory postLive = _liveBalancesAt(post);
+        uint256 preInvariant = _invariantOf(preLive, pre);
```
fredo: This recomputation is not independent verification of the swap math: the Vault obtains the swap result from `POOL.onSwap`, and this assertion then asks the same `POOL` implementation to evaluate `computeInvariant`. If those methods share a bug—or a custom pool deliberately makes them consistent—the assertion can pass while value leaks. For example, `onSwap` can return nearly the entire output balance while `computeInvariant` returns a constant; the invariant, supply, direction, and custody checks all pass. Please either restrict support to pinned trusted factory/codehash combinations and implement independent invariant math for each supported pool type, or narrow the claim to detecting inconsistencies between a pool’s own callbacks rather than broken/manipulated pool math.

examples/balancer/src/BalancerV3VaultAssertion.sol
```diff
@@ -0,0 +1,146 @@
...
+        registerTxEndTrigger(this.assertTokenRatesWithinDriftBound.selector);
+    }
+
+    /// @notice A swap must grow (or at worst preserve) the pool invariant and touch nothing else.
+    /// @dev The Vault takes `onSwap`'s output on trust: no `require` recomputes the curve. This
+    ///      check recomputes the pool's invariant from live balances through the pool's own math at
+    ///      the pre-call and post-call forks. With any nonzero swap fee the invariant must grow, so
+    ///      it may never drop by more than the configured rounding dust. BPT supply must be frozen
+    ///      across a swap, and live balances must move in the swap's direction (tokenIn in,
+    ///      tokenOut out); rates are fixed within one swap call, so live-balance direction mirrors
```
fredo: This assumption is incorrect for Balancer V3. A before-swap hook may change token rates, and the Vault explicitly reloads balances and rates before pricing the swap. Consequently, `preLive` and `postLive` can use different rate vectors even when the movement is allowed by `RATE_DRIFT_TOLERANCE_BPS`. A small valid rate decrease can make the live token-in balance fall despite its raw balance increasing, causing the direction or invariant check to reject an honest swap. Please use raw balances for direction and evaluate both invariant snapshots with one fixed rate vector, or require exact rate equality during the call instead of simultaneously allowing drift. Balancer documents and implements the rate reload here: https://github.com/balancer/balancer-v3-monorepo/blob/7861ea2785b96dd10681ff1b8dfe56b36cc202b6/pkg/vault/contracts/Vault.sol#L232-L240

foundry.toml
```diff
@@ -75,6 +75,14 @@ cache_path = "examples/aerodrome/cache"
 libs = ["lib"]
 remappings = ["credible-std/=src/"]
 
+[profile.balancer]
```
mateo-mro, line 78: The Balancer profile is not in the Solidity Test CI examples matrix, so none of these new tests run in PR CI. The checks are green today without executing `FOUNDRY_PROFILE=balancer pcl test`. Since this PR adds a new assertion suite, please add `balancer` to the examples matrix or state that this suite is only manually gated.

examples/balancer/src/BalancerV3VaultOutflowAssertion.sol
```diff
@@ -0,0 +1,64 @@
...
+    /// @notice Cumulative outflow cap as bps of the token balance at window start. 2000 = 20%.
+    uint256 public immutable outflowThresholdBps;
+
+    /// @notice Rolling window length, in seconds, over which the cap is enforced.
+    uint256 public immutable outflowWindowDuration;
+
+    constructor(address vault_, address token_, uint256 outflowThresholdBps_, uint256 outflowWindowDuration_) {
+        require(vault_ != address(0), "BalancerV3Outflow: zero vault");
+        require(token_ != address(0), "BalancerV3Outflow: zero token");
+        require(outflowThresholdBps_ != 0 && outflowThresholdBps_ <= 10_000, "BalancerV3Outflow: bad threshold");
```
fredo: Allowing `10_000` bps effectively disables this breaker. The executor dispatches only when net outflow strictly exceeds the threshold; a complete drain reaches exactly 100%, and net outflow cannot exceed the window-start balance. Therefore a full drain does not invoke the assertion at this allowed setting. Please require `< 10_000` and add an executor-backed regression test for an exact 100% drain. The strict-exceed comparison is here: https://github.com/phylaxsystems/credible-sdk/blob/832ecb884e4b2f80833262a6ad5cd93f2ed6d369/crates/assertion-executor/src/flow/state.rs#L238-L247

examples/balancer/src/BalancerV3VaultOutflowAssertion.sol
```diff
@@ -0,0 +1,64 @@
...
+/// @title BalancerV3VaultOutflowAssertion
+/// @author Phylax Systems
+/// @notice Rolling-window circuit breaker on one token leaving the Balancer V3 Vault's custody.
+/// @dev Apply to the Vault singleton. The Vault custodies every pool's liquidity for a token, so a
+///      drain of any single pool, buffer, or fee path ultimately shows up as that token leaving the
+///      Vault's ERC20 balance. This breaker is flow-based rather than selector-based: it does not
+///      care which Router, hook, or admin path moved the funds, only how fast custody is leaving,
+///      which is the one bound that survives a bug in any individual code path.
+///
+///      The breaker counts all exits, including honest large withdrawals and batch settlements, so
```
fredo: Not every token exit here is an economic drain. During a canonical ERC-4626 buffer wrap, underlying leaves the Vault while wrapped shares are received; a token-specific underlying watcher sees only the exit and can reject a value-preserving rebalance above the threshold. This needs explicit token-pair/value-aware handling or a documented restriction and calibration for buffer operations, rather than framing every observed exit as custody loss. Canonical buffer wrapping path: https://github.com/balancer/balancer-v3-monorepo/blob/7861ea2785b96dd10681ff1b8dfe56b36cc202b6/pkg/vault/contracts/Vault.sol#L1258-L1313

examples/balancer/src/BalancerV3VaultOutflowAssertion.sol
```diff
@@ -0,0 +1,64 @@
...
+    uint256 public immutable outflowThresholdBps;
+
+    /// @notice Rolling window length, in seconds, over which the cap is enforced.
+    uint256 public immutable outflowWindowDuration;
+
+    constructor(address vault_, address token_, uint256 outflowThresholdBps_, uint256 outflowWindowDuration_) {
+        require(vault_ != address(0), "BalancerV3Outflow: zero vault");
+        require(token_ != address(0), "BalancerV3Outflow: zero token");
+        require(outflowThresholdBps_ != 0 && outflowThresholdBps_ <= 10_000, "BalancerV3Outflow: bad threshold");
+        require(outflowWindowDuration_ != 0, "BalancerV3Outflow: zero window");
```
fredo: The executor rejects cumulative-flow windows shorter than its 10-second bucket and values that do not fit in `u64`, but this constructor accepts any nonzero `uint256`. The assertion can deploy successfully and then fail trigger registration/adoption. Please mirror the executor constraints here (`>= 10 seconds` and `<= type(uint64).max`) and cover registration in an integration test. Executor validation: https://github.com/phylaxsystems/credible-sdk/blob/832ecb884e4b2f80833262a6ad5cd93f2ed6d369/crates/assertion-executor/src/triggers/recorder.rs#L217-L225

examples/balancer/src/BalancerV3VaultOutflowAssertion.sol
```diff
@@ -0,0 +1,64 @@
...
+
+        registerAssertionSpec(AssertionSpec.Reshiram);
+    }
+
+    /// @notice Registers the rolling-window outflow breaker on the watched token.
+    /// @dev The executor tracks the rolling outflow internally and invokes the assertion only once
+    ///      the watched token's cumulative outflow crosses the threshold, so reaching the assertion
+    ///      already means the rate limit was breached.
+    function triggers() external view override {
+        watchCumulativeOutflow(
```
fredo, line 84: `watchCumulativeOutflow` thresholds net outflow (`cumulativeOutflow - cumulativeInflow`), not all exits as described above. Inflows from unrelated pools or buffers sharing the singleton can therefore offset a drain: for example, a 250-token deposit followed by a 400-token exit is only 150 tokens of measured outflow. If the intended protection is gross exits or a single-pool drain, this needs an absolute-outflow or pool-aware trigger. Reading `absoluteOutflow` inside the callback is insufficient because the callback is not dispatched when net flow remains below threshold. Executor calculation: https://github.com/phylaxsystems/credible-sdk/blob/832ecb884e4b2f80833262a6ad5cd93f2ed6d369/crates/assertion-executor/src/flow/state.rs#L198-L247

examples/balancer/src/BalancerV3VaultAssertion.sol
```diff
@@ -0,0 +1,146 @@
...
+
+        PhEvm.ForkId memory pre = _preCall(ctx.callStart);
+        PhEvm.ForkId memory post = _postCall(ctx.callEnd);
+
+        uint256[] memory preLive = _liveBalancesAt(pre);
+        uint256[] memory postLive = _liveBalancesAt(post);
+        uint256 preInvariant = _invariantOf(preLive, pre);
+        uint256 postInvariant = _invariantOf(postLive, post);
+        require(
+            postInvariant + _bpsOf(preInvariant, INVARIANT_DUST_TOLERANCE_BPS) >= preInvariant,
```
fredo: Basis points of the entire pool invariant cannot represent rounding dust safely. With an invariant of `1e24`, the smallest nonzero setting (`1` bp) permits a `1e20` invariant decrease per swap, while a zero setting can reject legitimate unit-level rounding. There is no configuration that allows a few rounding units without also allowing economically material, repeatable loss. Please use a small absolute bound derived from each supported pool implementation, optionally combined with a much finer relative tolerance or cumulative loss bound.

examples/balancer/src/BalancerV3VaultAssertion.sol
```diff
@@ -0,0 +1,146 @@
...
+            address rateProvider = snap.tokenInfo[i].rateProvider;
+            if (rateProvider == address(0)) {
+                continue;
+            }
+
+            uint256 postRate = _rateAt(rateProvider, postTx);
+            require(postRate != 0, "BalancerV3: rate provider returned zero rate");
+
+            uint256 preRate = _rateAt(rateProvider, preTx);
+            if (preRate == 0) {
```
fredo: A zero baseline must not be treated as equivalent to a provider that did not exist at the pre-transaction fork. An already deployed provider can successfully return zero before the transaction, change to an arbitrary nonzero rate before the Vault prices a swap, and then bypass the drift bound through this branch. Please preserve the read status and distinguish absent code / unavailable returndata from a successful encoded zero; a successful zero baseline should fail, and only a validated deployment-registration lifecycle should receive an exemption.

examples/balancer/src/BalancerV3VaultAssertion.sol
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        PoolTokenSnapshot memory snap = _poolTokenSnapshotAt(postTx);
+        for (uint256 i; i < snap.tokenInfo.length; ++i) {
+            address rateProvider = snap.tokenInfo[i].rateProvider;
+            if (rateProvider == address(0)) {
+                continue;
+            }
+
+            uint256 postRate = _rateAt(rateProvider, postTx);
```
fredo, line 278: This transaction-end callback runs for every Vault transaction, so this strict provider call makes the watched rate provider a dependency for recovery operations and unrelated pools sharing the singleton. Balancer recovery-mode exits deliberately use raw balances because external rate calls may fail; a reverting provider can therefore let the canonical recovery path succeed and then have this assertion revert the transaction. Please check post-state recovery mode before calling providers and scope the rate assertion to transactions involving the watched pool. Balancer documents the recovery constraint here: https://github.com/balancer/balancer-v3-monorepo/blob/7861ea2785b96dd10681ff1b8dfe56b36cc202b6/pkg/vault/contracts/VaultAdmin.sol#L394-L400

examples/balancer/src/BalancerV3VaultAssertion.sol
```diff
@@ -0,0 +1,146 @@
...
+            require(postRate != 0, "BalancerV3: rate provider returned zero rate");
+
+            uint256 preRate = _rateAt(rateProvider, preTx);
+            if (preRate == 0) {
+                // Provider only became live within this transaction (e.g. pool registration).
+                continue;
+            }
+
+            require(
```
fredo, line 289: Comparing only transaction endpoints does not prove that the rate stayed within this bound while the Vault used it. A before hook can raise the rate, the Vault can price the swap with that value, and an after hook can restore the original value; `preRate == postRate` then passes despite arbitrary transient manipulation. Balancer explicitly reloads hook-updated rates before swap pricing: https://github.com/balancer/balancer-v3-monorepo/blob/7861ea2785b96dd10681ff1b8dfe56b36cc202b6/pkg/vault/contracts/Vault.sol#L232-L240. This needs observation of the rate actually consumed by each Vault operation, or a restriction that forbids in-transaction provider changes.

examples/balancer/src/BalancerV3VaultOutflowAssertion.sol
```diff
@@ -0,0 +1,64 @@
...
+            token, outflowThresholdBps, outflowWindowDuration, this.assertOutflowWithinLimit.selector
+        );
+    }
+
+    /// @notice Hard circuit breaker for token outflow from the Vault.
+    /// @dev Invoked only after the cumulative outflow already exceeded the threshold, so it reverts
+    ///      unconditionally: the offending transaction is never included and the team can triage.
+    ///      A legitimate outflow above the cap must be split across windows or the limit raised.
```
fredo: Canonical aggregate-fee collection cannot necessarily be split across windows. `ProtocolFeeController.collectAggregateFees(pool)` has no amount parameter, and the Vault copies and zeros the complete accrued swap/yield fee ledger before transferring it. Once a legitimate fee balance exceeds this cap, waiting for another window does not reduce the amount, so every retry remains blocked. Please exempt a proven canonical collection bounded by the pre-transaction fee ledger, provide an operational bypass/cadence, or remove this guidance. Balancer collection path: https://github.com/balancer/balancer-v3-monorepo/blob/7861ea2785b96dd10681ff1b8dfe56b36cc202b6/pkg/vault/contracts/VaultAdmin.sol#L271-L297

examples/balancer/src/BalancerV3VaultOutflowAssertion.sol
```diff
@@ -0,0 +1,64 @@
...
+    function assertOutflowWithinLimit() external pure {
```
fredo: `vault` is stored but never checked against the actual assertion adopter. The flow watcher tracks the adopter, so adopting this assertion on the wrong account silently monitors and blocks that account while the configured Vault remains unprotected. Please make this callback `view` and verify `ph.getAssertionAdopter() == vault`, as the main assertion does, or remove the constructor parameter and document that the adopter is the sole source of truth. This should be covered through actual trigger registration and dispatch rather than a direct call to the unconditional revert.

examples/balancer/src/BalancerV3VaultAssertion.sol
makemake-kbo: Reworked in 822aeaf, together with the zero-baseline comment below. The pre-tx read now goes through a status-preserving helper (_rateStatusAt) instead of the strict decode, so a provider deployed inside the transaction no longer reverts before the exemption branch. The exemption itself is now keyed on pre-tx registration rather than a zero read. testProviderRegisteredDuringTxIsExempt deploys the provider within the armed transaction and passes.

examples/balancer/src/BalancerV3VaultAssertion.sol
makemake-kbo: Went with the restrict-and-validate option in 822aeaf: the swap check reads getHooksConfig at the pre-call fork and skips pools with shouldCallBeforeSwap or shouldCallAfterSwap. For the remaining pools no user code can run inside the swap call (the dynamic-fee hook is declared view, liquidity hooks do not run within swap), so the call-boundary snapshots do isolate the swap. The natspec and README state that hooked pools need a pool-specific variant. testHookedPoolSwapChecksAreSkipped covers the skip with a failure knob that would otherwise trip.

examples/balancer/src/BalancerV3VaultAssertion.sol
makemake-kbo: Took the third option in 822aeaf: renamed to assertPoolAccountingWithinVaultCustody and rewrote the docs to call it a necessary-but-not-sufficient per-pool bound. Aggregating every claim would need to enumerate all registered pools and buffers, which a one-pool example cannot do, and I did not find a pool-local conservation identity that holds across the shared ledgers. The shared-headroom case you describe is now explicitly documented as undetectable from one pool's view.

examples/balancer/src/BalancerV3VaultAssertion.sol
makemake-kbo: Narrowed the claim in 822aeaf. The docs now state this detects inconsistency between onSwap and the pool's own computeInvariant, i.e. the honest-but-buggy pool, and that a malicious implementation keeping both consistent passes. They also name the hardening path you describe (pinned factory/codehash combinations plus independent invariant math per supported pool type) as out of scope for the example.

examples/balancer/src/BalancerV3VaultAssertion.sol
makemake-kbo: Fixed in 822aeaf. Direction checks now use raw balances from getPoolTokenInfo. For rates: since pools with before/after-swap hooks are now skipped (thread above), nothing can legitimately move a rate inside a supported swap call, so the check requires exact rate equality at the two call boundaries instead of allowing drift within the call. That also fixes a single rate vector under both invariant evaluations. testRateShiftWithinSwapCallTrips covers the equality pin.

foundry.toml
makemake-kbo: Added balancer to the examples-pcl matrix in 822aeaf, so the suite is gated on every PR now.

examples/balancer/src/BalancerV3VaultOutflowAssertion.sol
makemake-kbo: Constructor now requires the threshold strictly below 10_000 (822aeaf), with the strict-exceed reasoning in a comment next to the guard. testRejectsFullDrainThreshold and testAcceptsMaximumEnforceableThreshold pin the boundary. On the executor-backed 100% drain regression: local pcl test does not simulate the rolling-window flow accounting (same limitation the fluid/mellow/lighter suites document), so the guard is the enforceable part here; the dispatch path is covered via a call-trigger subclass as in those suites.

examples/balancer/src/BalancerV3VaultOutflowAssertion.sol
makemake-kbo: Documented the restriction in 822aeaf rather than claiming value-awareness the trigger cannot provide. The contract docs and README now state the breaker is value-blind and token-scoped, that a canonical buffer wrap counts fully as outflow of the underlying, and that on deployments where the watched token backs an active buffer the cap must sit above the largest expected wrap (or the watched token should be one without buffer traffic).

examples/balancer/src/BalancerV3VaultOutflowAssertion.sol
makemake-kbo: Constructor now mirrors the executor bounds in 822aeaf: window must be >= the 10-second bucket and fit u64. Boundary tests cover 9, 10, u64 max, and u64 max + 1. Registration itself cannot be exercised by local pcl test (the flow trigger is executor-driven), so the guard makes the failure visible at deploy time instead.

examples/balancer/src/BalancerV3VaultOutflowAssertion.sol
makemake-kbo: Rewrote the docs in 822aeaf to state the net-flow semantics exactly: the threshold is on cumulativeOutflow minus cumulativeInflow, inflows from anything sharing the singleton offset exits, and a gross-exit bound would need an absolute-outflow or pool-aware trigger, which watchCumulativeOutflow is not. The framing of the breaker (contract natspec and README) now says sustained net custody loss rather than all exits.

examples/balancer/src/BalancerV3VaultAssertion.sol
makemake-kbo: Changed in 822aeaf: the tolerance is now an absolute value in invariant units instead of bps, with calibration guidance saying to derive it from the supported pool implementation's rounding behavior. The docs keep your point about why bps cannot work (one bp of a 1e24 invariant is 1e20 of repeatable loss). I left it as a single absolute bound rather than adding a combined relative/cumulative term, since a per-swap dust bound is the part a call-scoped check can enforce.

examples/balancer/src/BalancerV3VaultAssertion.sol
makemake-kbo: Fixed in 822aeaf. The baseline read now preserves status: registered-pre-tx providers must answer with a nonzero pre-tx rate, and a successful encoded zero reverts instead of taking the exemption (testZeroBaselineForRegisteredProviderTrips). The exemption is granted only to providers absent from the pre-tx pool registration, which is the validated lifecycle signal: token config is immutable after pool registration, so a new provider for the watched pool can only appear when the pool itself is registered within the transaction.

examples/balancer/src/BalancerV3VaultAssertion.sol
makemake-kbo: Both gates added in 822aeaf. The rate check returns early when isPoolInRecoveryMode(POOL) is true post-tx, and it only runs when the transaction changed the watched pool's accounting (raw balances, BPT supply, or aggregate fees; live balances deliberately excluded since they move with the rate itself). Unrelated singleton traffic never calls the provider anymore. The residual is documented on the assertion: a transaction that moves a rate without touching the pool is not examined, but any later transaction consuming that rate touches the pool and is. Tests: testRecoveryModeSkipsRateAssertion, testRateOnlyTransactionIsOutOfScope.

examples/balancer/src/BalancerV3VaultAssertion.sol
makemake-kbo: Addressed in 822aeaf in two parts. Hooked pools are skipped by the swap check, so the before-hook manipulation you describe cannot happen on a supported pool: nothing can move a rate inside a hookless swap call, and that is enforced with an exact equality pin at the call boundaries. For manipulation between calls of one transaction, the swap check now also bounds the swap-start rate (the value the Vault prices against) to the pre-tx baseline. testTransientRateManipulationAroundSwapTrips exercises a manipulate-swap-restore router and trips at the swap; testTransientRateManipulationPassesEndpointDrift documents that endpoint comparison alone misses it.

examples/balancer/src/BalancerV3VaultOutflowAssertion.sol
makemake-kbo: Removed the split-across-windows guidance in 822aeaf. The docs now say collectAggregateFees transfers the whole accrued ledger in one indivisible call, that collection cadence has to keep the ledger below the cap, and that an operation that has already outgrown the cap can only proceed by raising the limit or temporarily deactivating the assertion through the Credible Layer management flow. I did not build a proven-canonical-collection exemption into the callback, since the breaker only dispatches after the threshold is already breached and the exemption would have to re-derive the whole window from inside one transaction.

examples/balancer/src/BalancerV3VaultOutflowAssertion.sol
makemake-kbo: Done in 822aeaf: the callback is view and requires ph.getAssertionAdopter() == vault with its own error message, so misadoption fails loudly instead of rate-limiting the wrong account. Dispatch is covered through real trigger registration via a call-trigger subclass (the flow trigger itself is executor-driven and not simulated by local pcl test, same as the other flow-breaker suites): testDispatchedBreakerReverts and testDispatchedBreakerRejectsWrongAdopter, which pin the two distinct messages. The direct-call test is gone.

examples/balancer/src/BalancerV3VaultAssertion.sol
```diff
@@ -0,0 +1,277 @@
...
+        PhEvm.ForkId memory pre = _preCall(ctx.callStart);
+        PhEvm.ForkId memory post = _postCall(ctx.callEnd);
+
+        PoolTokenSnapshot memory preSnap = _poolTokenSnapshotAt(pre);
+
+        _requireSwapRatesWithinBaseline(preSnap, pre);
+
+        uint256[] memory preLiveBalances = _liveBalancesAt(pre);
+        uint256[] memory postLiveBalances = _liveBalancesAt(post);
+        uint256 preInvariant = _invariantOf(preLiveBalances, pre);
```
mateo-mro, line 120: The BPT supply check is gone, so a `swap` that performs an otherwise normal balance update and also mints BPT now passes. The removed `SupplyDrift` fixture covered exactly that case. `swap` should never mint or burn pool BPT, and the live balance plus invariant checks do not constrain supply, so this silently drops one of the advertised protections. Please retain an equivalent guard and regression test.

mateo-mro (changes requested):

examples/balancer/src/BalancerV3VaultAssertion.sol
```diff
@@ -0,0 +1,280 @@
...
+        registerAssertionSpec(AssertionSpec.Reshiram);
+    }
+
+    /// @notice Registers Vault selectors against their protection assertions.
+    /// @dev The Vault is the assertion adopter. The swap check is call-scoped so it compares the
+    ///      exact pre-call and post-call snapshots of the matched swap; the custody and rate checks
+    ///      are transaction-wide envelopes because add/remove liquidity, buffer operations, and fee
+    ///      collection all move the same accounting.
+    function triggers() external view override {
+        registerFnCallTrigger(this.assertSwapPreservesPoolInvariant.selector, IBalancerV3VaultLike.swap.selector);
```
mateo-mro, line 53: Only `swap` gets a call scoped rate observation. Balancer also consumes `poolData.tokenRates` in `addLiquidity` and `removeLiquidity`, so a router can move a provider, execute an unbalanced liquidity operation, then restore the rate before transaction end. The endpoint check sees the original rate and passes even though the pool priced the operation at the manipulated rate. Please cover those selectors with a scoped check or make this limitation explicit.

examples/balancer/src/BalancerV3VaultAssertion.sol
```diff
@@ -0,0 +1,280 @@
...
+        registerTxEndTrigger(this.assertPoolAccountingWithinVaultCustody.selector);
```
mateo-mro, line 66: These callbacks execute after every Vault transaction, including traffic for unrelated pools. For an eight token pool, an unrelated transaction performs 73 snapshot reads before the rate callback establishes that watched accounting did not change: 34 in the custody check and 39 in the rate check. That is a significant cost on the singleton path and can exhaust the assertion budget for traffic this suite intends to ignore. Please gate the reads on calls that touch the watched pool, or split registration by the relevant selectors.

examples/balancer/src/BalancerV3VaultAssertion.sol
```diff
@@ -0,0 +1,357 @@
...
+
+    /// @dev Cheap tx-end gate. `getCallInputs` reads the trace without any Vault snapshots, so
+    ///      unrelated singleton traffic returns before the per-token accounting and provider work.
+    function _transactionTouchesWatchedPool() internal view returns (bool) {
+        return _structSelectorTouchesWatchedPool(IBalancerV3VaultLike.swap.selector, 32)
+            || _structSelectorTouchesWatchedPool(IBalancerV3VaultLike.addLiquidity.selector, 0)
+            || _structSelectorTouchesWatchedPool(IBalancerV3VaultLike.removeLiquidity.selector, 0)
+            || _addressSelectorTouchesWatchedPool(IBalancerV3VaultLike.initialize.selector)
+            || _addressSelectorTouchesWatchedPool(IBalancerV3VaultLike.removeLiquidityRecovery.selector)
+            || _addressSelectorTouchesWatchedPool(IBalancerV3VaultLike.collectAggregateFees.selector)
```
mateo-mro, line 333: `collectAggregateFees` only clears aggregate fee balances and supplies raw token credit. It never loads pool data or reads a rate provider. Including it in this shared gate makes `assertTokenRatesWithinDriftBound` block fee collection when a provider is broken or out of bounds, although the Vault itself would complete the call. Please separate the custody gate from the rate gate and leave fee collection out of the latter.

examples/balancer/test/BalancerV3VaultAssertion.t.sol
```diff
@@ -0,0 +1,319 @@
...
+        _arm(BalancerV3VaultAssertion.assertTokenRatesWithinDriftBound.selector);
+        vm.expectRevert(bytes("BalancerV3: zero rate baseline"));
+        _swap(address(pool));
+    }
+
+    /// @notice A provider deployed and registered within the transaction has no pre-tx baseline by
+    ///         construction: the deployment lifecycle is exempt from the drift comparison (only the
+    ///         nonzero post-state is enforced) instead of reverting on the missing baseline read.
+    function testProviderRegisteredDuringTxIsExempt() public {
+        vault.registerNewRateProvider();
```
mateo-mro, line 254: `registerNewRateProvider` runs before the armed `initialize` call, so the provider already exists at the pre transaction fork and this never reaches the missing baseline branch. The old helper performed the deployment inside the watched call. Please restore that shape so the registration exemption is actually covered.

mateo-mro (approved):

examples/balancer/src/BalancerV3VaultOutflowAssertion.sol
makemake-kbo: Addressed in 822aeaf by documenting the operational reality rather than claiming an exemption the trigger can't provide. `watchCumulativeOutflow` is executor-driven and can't special-case `collectAggregateFees`, so the natspec now states the collection moves the entire accrued fee ledger in one call (no amount parameter, not splittable across windows), that collection cadence must keep the accrued ledger under the cap, and that a legitimate collection which has already outgrown the cap can only proceed by raising the limit or temporarily deactivating this assertion through the Credible Layer management flow.

examples/balancer/src/BalancerV3VaultOutflowAssertion.sol
makemake-kbo: Fixed in 822aeaf. `assertOutflowWithinLimit` is now `view` and requires `ph.getAssertionAdopter() == vault` before the unconditional revert, so adopting on the wrong account fails loudly instead of silently rate-limiting it while the Vault goes unprotected — the same adopter-is-source-of-truth check the main assertion uses. Dispatch is exercised through a call-trigger subclass (`ArmedOutflowBreaker` registers a call trigger on the unmodified selector; `triggers()` is virtual for exactly this): `testDispatchedBreakerReverts` covers the matching-adopter revert and `testDispatchedBreakerRejectsWrongAdopter` covers the adopter mismatch, both through real dispatch rather than a direct call to the revert.

PR #88, opened 2026-07-20
fix: repair protocol assertion semantics

makemake-kbo:
## Summary

This repairs protocol assertion semantics found during an audit of `credible-std` at `a98a1ba2`.

- replace shared ERC-4626 all-snapshot comparisons with endpoint checks and validate deployment configuration
- fix version-specific ABI, arithmetic, and state-transition conflicts across the protocol adapters
- narrow or quarantine checks that depend on unsupported policy assumptions, noncausal evidence, or incompatible protocol models
- add regression coverage and update protocol documentation to match the active protection surface

## Root cause

Several assertions treated transient execution states, events, return values, idle-token flow, or stricter policy assumptions as protocol invariants. Other assertions depended on stale selectors, layouts, or executor-invalid configuration.

## Impact

The repaired paths no longer reject the protocol-valid transitions covered by the audit witnesses. Assertions that need a new ABI or economic model are visibly quarantined instead of presenting an unsupported release guarantee.

Some constructor ABIs and trigger surfaces change. Existing deployments need updated configuration and should not activate quarantined bundles until protocol-specific replacements exist.

## Validation

- all 22 Foundry profiles build offline
- all 21 protocol profiles pass with PCL 1.5.1
- root protection suite passes: 131 tests
- `forge fmt --check`
- `git diff --check`

## Notes

The separate `makemake/eng-3875-create-assertions-for-balancer` branch was reviewed but is not included here. It adds a Balancer suite and does not modify the assertions repaired by this PR.


examples/tydro/src/TydroOperationSafety.sol
```diff
@@ -13,6 +14,28 @@ interface ITydroL2Pool {
     function setUserUseReserveAsCollateral(bytes32 args) external;
 }
 
+interface ITydroPoolCurrent {
+    function finalizeTransfer(
```
mateo-mro, line 20: Tydro’s `finalizeTransfer` has six arguments, with `amount` between `to` and the two balances. This declaration derives a different selector, so actual aToken callbacks are not monitored or decoded and collateral transfers bypass the health check. Please use the standard signature and decode the amount.

examples/tydro/src/TydroOperationSafety.sol
```diff
@@ -90,16 +116,16 @@ contract TydroProtectionSuite is AaveV3LikeProtectionSuite {
             operation.asset = _assetByL2Id(args1 >> 16);
             operation.relatedAsset = _assetByL2Id(args1);
             operation.counterparty = triggered.caller;
-            operation.amount = _decodeL2Amount(args2);
-            operation.metadata = abi.encode(((uint256(args2) >> 128) & 1) == 0);
+            operation.amount = uint256(args2) & L2_SHORTENED_AMOUNT_MASK;
```
mateo-mro, line 125: The L2 liquidation sentinel is part of this compact amount encoding. `type(uint128).max` means `type(uint256).max`, but this mask leaves it shortened. Reuse `_decodeL2Amount` or expand the sentinel here so the operation context matches the call.

examples/tydro/src/TydroOperationSafety.sol
```diff
@@ -90,16 +116,16 @@ contract TydroProtectionSuite is AaveV3LikeProtectionSuite {
...
+            operation.metadata = abi.encode(((uint256(args2) >> 128) & 1) != 0);
```
mateo-mro: The L2 encoding defines zero as aToken settlement and one as underlying settlement. This is inverted here. `_getLiquidationCollateralCheck` uses this flag to choose both the token and sender, so it can inspect a zero transfer while collateral is actually seized. Store `== 0` and cover both values.

examples/tydro/src/TydroOperationSafety.sol
```diff
@@ -90,16 +116,16 @@ contract TydroProtectionSuite is AaveV3LikeProtectionSuite {
...
             return operation;
         }
 
         if (triggered.selector == ITydroL2Pool.setUserUseReserveAsCollateral.selector) {
             bytes32 args = abi.decode(triggered.input[4:], (bytes32));
-            bool disableCollateral = ((uint256(args) >> 16) & 1) != 0;
+            bool useAsCollateral = ((uint256(args) >> 16) & 1) != 0;
```
mateo-mro: Bit 16 has the opposite meaning in L2Pool: zero enables collateral and one disables it. This decodes an enable as `DisableCollateral` and lets a real disable call bypass the post operation solvency check. Please flip the comparison and cover both encoded values.

mateo-mro (changes requested):

examples/tydro/src/TydroOperationSafety.sol
```diff
@@ -65,7 +94,7 @@ contract TydroProtectionSuite is AaveV3LikeProtectionSuite {
             operation.kind = OperationKind.Borrow;
             operation.account = triggered.caller;
             operation.asset = _assetByL2Id(args);
-            operation.amount = _decodeL2Amount(args);
+            operation.amount = _decodeL2Amount(args, false);
```
mateo-mro: `borrow` has the same max amount sentinel as the other L2 methods. The current check only cares whether the amount is zero, but passing `false` leaves the decoder reporting `uint128` max for a call the Pool treats as `uint256` max. Please expand it and add the matching regression case.

mateo-mro (dismissed):

examples/tydro/src/TydroOperationSafety.sol
makemake-kbo: Fixed in dd56b28. The L2 `borrow` branch now decodes with the sentinel expanded (`_decodeL2Amount(args, true)`), matching `withdraw` and the Pool's own `type(uint128).max` → `type(uint256).max` handling, so the operation amount is no longer reported shortened. `testL2BorrowDecodesMaxAmountSentinel` packs a max-sentinel borrow and asserts the expanded `uint256` max.

examples/symbiotic/src/SymbioticVaultFlowAssertion.sol
```diff
@@ -143,17 +165,60 @@ abstract contract SymbioticVaultFlowAssertion is SymbioticVaultFlowHelpers {
         require(deltas.vaultOutflow == claimedAmount, "SymbioticVault: claimBatch vault outflow mismatch");
         require(deltas.recipientInflow == claimedAmount, "SymbioticVault: claimBatch recipient inflow mismatch");
 
+        uint256 entitlement;
         for (uint256 i; i < epochs.length; ++i) {
             require(epochs[i] < currentEpoch, "SymbioticVault: claimBatch succeeded for immature epoch");
-            _assertClaimStateTransition(epochs[i], call_.caller, preFork, postFork, true);
+            entitlement += _claimEntitlementAt(epochs[i], caller, preFork);
```
mateo-mro, line 177: Please reject repeated epoch values before adding the entitlement. With `[e, e]`, both transition checks inspect the same pre and post state and pass, while `entitlement` counts the claim twice. A faulty batch implementation that pays the same epoch twice would satisfy this assertion.

examples/symbiotic/src/SymbioticVaultFlowAssertion.sol
```diff
@@ -26,17 +26,20 @@ abstract contract SymbioticVaultFlowAssertion is SymbioticVaultFlowHelpers {
         registerFnCallTrigger(this.assertRedeemScheduling.selector, ISymbioticVaultLike.redeem.selector);
         registerFnCallTrigger(this.assertClaimFlow.selector, ISymbioticVaultLike.claim.selector);
         registerFnCallTrigger(this.assertClaimBatchFlow.selector, ISymbioticVaultLike.claimBatch.selector);
-        registerTxEndTrigger(this.assertTotalStakeIdentity.selector);
+        registerFnCallTrigger(this.assertSlashAccounting.selector, ISymbioticVaultLike.onSlash.selector);
+        // `totalStake()` is defined from these same bucket getters in v1, so a transaction-end
```
mateo-mro, line 30: Could we update the bundle description too? It still says this bundle protects the transaction wide total stake identity, but this change no longer registers that assertion.

src/protection/swaps/examples/CowSettlementAssertion.sol
```diff
@@ -94,18 +40,13 @@ contract CowSettlementAssertion is CowSettlementHelpers {
...
-            uint256 settlementVolume = _reportedTradeVolume(token);
-            uint256 tolerance = (pre * BUFFER_TOLERANCE_BPS) / BPS;
-            uint256 allowedOut = _saturatingAdd(_saturatingAdd(swept, settlementVolume), tolerance);
-
-            require(grossOut <= allowedOut, "CowSettlement: buffer moved to unauthorized recipient");
-
-            if (post < pre) {
-                uint256 netOut = pre - post;
-                require(netOut <= allowedOut, "CowSettlement: buffer drained to unauthorized recipient");
+            if (post >= pre || _settlementExecutionOccurred()) {
```
mateo-mro: The endpoint check can be bypassed by prefunding the Settlement in the same transaction. With a standing allowance, the attacker can transfer `D` in, pull the original balance plus `D` to themselves, and finish with `post == pre`. The branch then skips although the original buffer was stolen. This needs to account for gross outflow, not only the net endpoint.

src/protection/swaps/examples/CowSettlementHelpers.sol
```diff
@@ -66,83 +54,8 @@ abstract contract CowSettlementHelpers is Assertion {
...
-    }
-
-    /// @notice Saturating sum for small allowance sets used in reconciliation.
-    function _saturatingAdd(uint256 a, uint256 b) internal pure returns (uint256) {
-        unchecked {
-            uint256 c = a + b;
-            return c < a ? type(uint256).max : c;
-        }
+    function _settlementExecutionOccurred() internal view returns (bool) {
+        return ph.getCallInputs(SETTLEMENT, IGPv2SettlementLike.settle.selector).length != 0
```
mateo-mro: `settle` accepts arbitrary interactions. A compromised solver can make a valid settlement and use one of those interactions to consume the standing allowance, but this transaction wide selector check suppresses the destination check for every watched token. The exception needs to be tied to the particular token outflow, otherwise the same drain passes.

src/protection/swaps/examples/UniswapV4PoolManagerAssertion.sol
```diff
@@ -74,11 +74,16 @@ contract UniswapV4PoolManagerAssertion is UniswapV4PoolManagerHelpers {
         if (!_matchesConfiguredPool(key)) {
             return;
         }
+        if (key.hooks != address(0)) {
```
mateo-mro: Any configured key with a nonzero hook now makes both call scoped protections return without checking price movement or liquidity, although construction and trigger registration still succeed. That leaves the deployment looking active while swaps and liquidity changes are unchecked. Reject hooked pools at construction until the hook path has an assertion model.

mateo-mro (changes requested):

src/protection/swaps/examples/CowSettlementAssertion.sol
```diff
@@ -44,68 +24,25 @@ contract CowSettlementAssertion is CowSettlementHelpers {
...
-            uint256 tolerance = (pre * BUFFER_TOLERANCE_BPS) / BPS;
-            uint256 allowedOut = _saturatingAdd(_saturatingAdd(swept, settlementVolume), tolerance);
-
-            require(grossOut <= allowedOut, "CowSettlement: buffer moved to unauthorized recipient");
 
-            if (post < pre) {
-                uint256 netOut = pre - post;
-                require(netOut <= allowedOut, "CowSettlement: buffer drained to unauthorized recipient");
-            }
+            require(grossOut <= swept + settlementVolume, "CowSettlement: external buffer drain");
```
mateo-mro: `swept` and `settlementVolume` are not disjoint. If the configured Safe is the receiver of a valid trade for this token, that transfer is counted once as trade volume and again as a sweep. An additional unauthorized outflow of the same amount then still satisfies this check. Please separate the two allowances or reject transactions where their attribution can overlap.

mateo-mro (changes requested):

src/protection/swaps/examples/CowSettlementAssertion.sol
```diff
@@ -44,68 +26,26 @@ contract CowSettlementAssertion is CowSettlementHelpers {
...
-            uint256 tolerance = (pre * BUFFER_TOLERANCE_BPS) / BPS;
-            uint256 allowedOut = _saturatingAdd(_saturatingAdd(swept, settlementVolume), tolerance);
-
-            require(grossOut <= allowedOut, "CowSettlement: buffer moved to unauthorized recipient");
 
-            if (post < pre) {
-                uint256 netOut = pre - post;
-                require(netOut <= allowedOut, "CowSettlement: buffer drained to unauthorized recipient");
-            }
+            uint256 authorizedOutflow = swept > settlementVolume ? swept : settlementVolume;
```
mateo-mro: This now rejects a valid transaction that contains both ordinary trade outflow and a separate Safe sweep. With trade volume `V` and a distinct sweep `S`, `grossOut` is `V + S` while this authorizes only `max(V, S)`. Since `settle` can execute arbitrary interactions, those two outflows can occur together. Please model the disjoint case or explicitly quarantine that unsupported combination.

examples/symbiotic/src/SymbioticVaultFlowAssertion.sol
makemake-kbo: Fixed in c54a3f2. `assertClaimBatchFlow` now rejects any repeated epoch value (`SymbioticVault: claimBatch duplicate epoch`) before accumulating the entitlement, so a faulty batch implementation paying the same epoch twice can no longer satisfy the per-epoch transition checks. `testClaimBatchRejectsRepeatedEpoch` drives a mock that pays `[e, e]` twice and asserts the trip; `testClaimBatchPaysEachMatureEpochOnce` covers the honest multi-epoch batch.

examples/symbiotic/src/SymbioticVaultFlowAssertion.sol
makemake-kbo: Fixed in c54a3f2. The bundle description now lists the surface it actually wires (deposit, withdraw, redeem, claim, claimBatch, slashing) and notes why `assertTotalStakeIdentity` is intentionally not registered for the v1 vault.

src/protection/swaps/examples/CowSettlementAssertion.sol
makemake-kbo: Fixed in 8e07a46. The allowances are now disjoint by destination: Safe-bound outflow is always an authorized destination, and outflow to anyone else must fit the token's reported trade volume, so an honest settlement whose trades pay ordinary receivers no longer competes with a sweep for the same allowance. The one remaining overlap — both destinations paid more than trade volume explains — is observationally identical to "trade paid the Safe plus an equal-sized drain" (Trade events carry no receiver), so it is explicitly quarantined with its own error (`CowSettlement: unattributable sweep and trade outflow`) rather than credited both allowances. `testTradeVolumePlusSeparateSweepIsQuarantined` documents that known false positive, and the earlier drain regressions still trip.

examples/tydro/src/TydroOperationSafety.sol
makemake-kbo: Fixed in 772aa86. The interface now declares the standard six-argument `finalizeTransfer(asset, from, to, amount, scaledBalanceFromBefore, scaledBalanceToBefore)`, so the derived selector matches the real aToken callback, and the decoder reads the transferred amount. Covered by `testFinalizeTransferUsesCanonicalSelectorAndAmount`.

examples/tydro/src/TydroOperationSafety.sol
makemake-kbo: Fixed in 772aa86. The L2 liquidation decode now applies the shortened-amount mask and expands the `type(uint128).max` sentinel to `type(uint256).max`, matching `_decodeL2Amount` semantics, so the operation context reports the same amount the Pool acts on.

examples/tydro/src/TydroOperationSafety.sol
makemake-kbo: Fixed in 772aa86. The receiveAToken flag is now stored as `== 0` (zero = aToken settlement, one = underlying), so `_getLiquidationCollateralCheck` inspects the token and sender that actually settle the seized collateral. Regression tests cover both encoded values.

src/protection/swaps/examples/CowSettlementAssertion.sol
makemake-kbo: Fixed in 0fc0cef. The check now reconciles gross outflow (summed `Transfer`s out of the Settlement) instead of the endpoint balance delta, so same-transaction prefunding no longer hides a standing-allowance drain. `testSameTransactionPrefundingDoesNotHideGrossOutflow` reproduces the prefund-then-pull sequence and asserts the trip.

src/protection/swaps/examples/CowSettlementHelpers.sol
makemake-kbo: Fixed in 0fc0cef. The transaction-wide `settle`/`swap` selector exception is gone; a settlement call now only earns each token an allowance equal to that token's own reported `Trade` volume, so an interaction consuming a standing allowance is no longer excused by the presence of a valid settlement. `testSettlementCallOnlyAuthorizesReportedTokenVolume` covers the compromised-solver drain.

src/protection/swaps/examples/UniswapV4PoolManagerAssertion.sol
makemake-kbo: Fixed in 0fc0cef. Hooked pool keys are now rejected at construction until the hook path has an assertion model, so a deployment can no longer look active while both call-scoped protections silently return. Covered by `testRejectsHookedPoolWithoutHookAssertionModel`.

src/protection/swaps/examples/CowSettlementAssertion.sol
makemake-kbo: Addressed in 39fb46b and reworked in 8e07a46 after the follow-up: the allowances are now disjoint by destination (Safe-bound outflow vs trade-volume-bounded outflow to everyone else), and the case where attribution can overlap is explicitly quarantined. `testTradeToSweepSafeCannotAuthorizeAdditionalDrain` keeps the equal-sized-drain scenario tripping.

examples/balancer/src/BalancerV3VaultAssertion.sol
```diff
@@ -196,17 +199,31 @@ contract BalancerV3VaultAssertion is BalancerV3VaultHelpers {
...
-        if (!_transactionTouchesWatchedPool()) {
-            return;
-        }
         PhEvm.ForkId memory post = _postTx();
         if (!_isPoolInitializedAt(post)) {
             return;
         }
 
         PoolTokenSnapshot memory snap = _poolTokenSnapshotAt(post);
+        if (!_watchedPoolAccountingChangedTxWide(snap, post)) {
```
mateo-mro: This gate skips the failure the assertion says it catches. A buggy watched pool call can transfer custody out while leaving reserves, raw balances, supply, and fees unchanged, so this returns and accepts the draining transaction. Detecting the deficit on a later pool operation is too late for a transaction blocking assertion. Please use a bounded gate that includes custody movement, or remove the claim that this prevents unrecorded custody loss.

mateo-mro (changes requested):

examples/balancer/src/BalancerV3VaultAssertion.sol
makemake-kbo: Fixed in 704197b. The accounting-delta gate is gone. The check now computes each bound as a per-token deficit at both transaction endpoints and rejects the transaction only when a deficit grew, so a transaction that moves custody without touching the pool's raw balances, supply, or fees is blocked at the causing transaction. Cost stays bounded by the pool's token count (storage reads plus balanceOf, no trace scans), and a deficit that predates the transaction still doesn't block honest traffic. `testCustodyDrainWithoutAccountingChangeTrips` covers your scenario; `testWorseningPreexistingDeficitTrips` and `testHonestOperationOnDeficientPoolPasses` pin the comparison semantics.

examples/balancer/src/BalancerV3VaultAssertion.sol
```diff
@@ -196,29 +206,64 @@ contract BalancerV3VaultAssertion is BalancerV3VaultHelpers {
...
         if (!_isPoolInitializedAt(post)) {
             return;
         }
 
+        PhEvm.ForkId memory pre = _preTx();
         PoolTokenSnapshot memory snap = _poolTokenSnapshotAt(post);
         for (uint256 i; i < snap.tokens.length; ++i) {
             address token = snap.tokens[i];
             uint256 reserves = _reservesOfAt(token, post);
+            uint256 custodyDeficit = _deficit(reserves, _readBalanceAt(token, VAULT, post));
```
mateo-mro, line 233: This makes every Vault transaction depend on `balanceOf` for every watched token, even when the transaction only touches an unrelated pool. A paused or upgraded token that starts reverting can therefore make the entire singleton unusable, while the Vault could still process other pools. The residual note only explains why the watched pool cannot operate. Please gate the read on a bounded watched token change signal or explicitly quarantine the singleton wide liveness impact.

mateo-mro (changes requested):

mateo-mro (dismissed):

mateo-mro (approved):

PR #89, opened 2026-07-21
fix(vault): bound the ERC4626 tx-wide share-price envelope to avoid PrecompileOOG

makemake-kbo:
## Problem

The `metamorpho-top-5-vaults` deployment produced repeated **false-positive invalidations**. Backtesting against the assertion-replay service pinned the cause to the tx-wide share-price envelope in `ERC4626SharePriceAssertion`.

`assertSharePriceEnvelope()` is registered with `registerTxEndTrigger`, so it fires on **every** transaction that merely touches the vault — including read-only calls and unrelated protocol activity. Its first check, `ph.assetsMatchSharePrice(vault, tol)`, compares the pre-tx share price against **every** fork point in the transaction. Fork enumeration is `PostTx + PreCall(i) + PostCall(i)` per forkable call, so the vault's `totalAssets()`/`totalSupply()` are re-read at ~`1 + 2*callFrames` points.

MetaMorpho's `totalAssets()` is not a storage read — it loops the Morpho Blue supply queue (per-market interest accrual). On a high-call-count transaction (e.g. a 107-call AVS task submission that merely touches the vault), the envelope re-runs that computation ~215 times and exhausts the assertion gas limit → **PrecompileOOG → revert → false invalidation**. The share price itself was flat/rising across the offending blocks; there was no real violation.

## Fix

Add `assertSharePriceEnvelopeBounded()` — a tx-wide envelope evaluated at only the **pre-tx and post-tx** forks via `assetsMatchSharePriceAt` (2 fork points, O(1) in call-frame count, so it cannot OOG on high-call-count txs).

The precompile's deviation check is symmetric (`abs_diff`), so the bounded envelope is **two-sided**: it still catches both dilution (share price down) and unexpected inflation (share price up — e.g. a donation / direct-transfer manipulation) reached through **any** entrypoint, including non-ERC-4626 selectors the per-call check never observes. The per-call checks continue to cover settlement during standard `deposit`/`mint`/`withdraw`/`redeem` ops.

Trigger registration is split into named helpers (no conditional flag):

- `_registerSharePriceTriggers()` — keeps the exhaustive all-forks envelope (**unchanged default**; existing consumers such as the Curve/LlamaLend and Spark examples are unaffected)
- `_registerBoundedSharePriceTriggers()` — bounded envelope + per-call
- `_registerPerCallSharePriceTriggers()` — shared per-call registration

`MetaMorphoVaultAssertion.triggers()` now uses `_registerBoundedSharePriceTriggers()`.

## Trade-off

The bounded envelope does not inspect intermediate call-boundary forks, so a share-price dip that fully recovers *within* the same transaction via a non-standard entrypoint is no longer flagged tx-wide. Settlement during standard ERC-4626 ops is still covered per-call, and the net pre→post move (in both directions) is still enforced. For computed-`totalAssets()` vaults the exhaustive scan is not usable regardless.

## Verification

Replayed against the assertion-replay service over the previously-failing window (`25480001 + 560`). With the bounded bundle, both **Gauntlet USDC Prime** (`0xdd0f28…`) and **Steakhouse USDC** (`0xBEEF01…`) scan the **entire** block range with **no invalidation** — including the exact high-call-count transactions (Gauntlet block 25480541, Steakhouse 25480212) that previously OOG'd the envelope. The bounded envelope fires on those txs (it is a txEnd trigger) and completes within budget.

> Note: no local `forge`/`pcl test` regression is added — the reshiram share-price precompiles and fork snapshots aren't exercised by the local test runner, so validation is via the replay service (the environment these assertions actually run in).

### DIFF
--- a/src/protection/vault/ERC4626SharePriceAssertion.sol
+++ b/src/protection/vault/ERC4626SharePriceAssertion.sol
@@ -30,14 +30,35 @@ abstract contract ERC4626SharePriceAssertion is ERC4626BaseAssertion {
         sharePriceToleranceBps = _toleranceBps;
     }
 
-    /// @notice Register the default trigger set for share-price invariants.
-    /// @dev Uses registerTxEndTrigger for the tx-wide envelope and registerFnCallTrigger
-    ///      for per-call checks. Call this inside your `triggers()`.
+    /// @notice Register the default trigger set: the exhaustive tx-wide envelope + per-call checks.
+    /// @dev The tx-wide envelope (`assertSharePriceEnvelope`) uses the all-fork-points
+    ///      `assetsMatchSharePrice` scan, which re-reads `totalAssets()`/`totalSupply()` at
+    ///      ~1+2*callFrames fork points. This is fine for vaults with a cheap `totalAssets()`.
+    ///      Vaults with an expensive computed `totalAssets()` (e.g. MetaMorpho, which loops the
+    ///      Morpho Blue supply queue) should instead use `_registerBoundedSharePriceTriggers()`:
+    ///      the exhaustive scan can exhaust the assertion gas limit on unrelated high-call-count
+    ///      transactions that merely touch the vault and revert (PrecompileOOG) — a false invalidation.
     function _registerSharePriceTriggers() internal view {
-        // Tx-wide envelope — fires once after the transaction completes
         registerTxEndTrigger(this.assertSharePriceEnvelope.selector);
+        _registerPerCallSharePriceTriggers();
+    }
 
-        // Per-call check — fires once per matching ERC-4626 operation
+    /// @notice Register a gas-bounded trigger set: a pre/post tx-wide envelope + per-call checks.
+    /// @dev The tx-wide check (`assertSharePriceEnvelopeBounded`) compares share price only between
+    ///      the pre-tx and post-tx forks — 2 fork points, O(1) in call-frame count — so it never
+    ///      triggers the all-forks scan that OOGs on expensive-`totalAssets()` vaults. It is
+    ///      two-sided: it catches both dilution (share price down) and unexpected inflation (share
+    ///      price up, e.g. a donation / direct-transfer manipulation) reached through ANY entrypoint,
+    ///      including non-ERC-4626 selectors the per-call check never observes.
+    function _registerBoundedSharePriceTriggers() internal view {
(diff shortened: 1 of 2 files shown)


src/protection/vault/ERC4626SharePriceAssertion.sol
```diff
@@ -69,6 +90,20 @@ abstract contract ERC4626SharePriceAssertion is ERC4626BaseAssertion {
         );
     }
 
+    /// @notice Tx-wide share-price envelope bounded to the pre-tx vs post-tx comparison.
+    /// @dev Two-sided within tolerance — catches both dilution (decrease) and unexpected inflation
+    ///      (increase, e.g. a donation / direct-transfer share-price manipulation), regardless of the
```
mateo-mro, line 95: `assetsMatchSharePriceAt` returns true when either fork has zero shares. A first deposit, final redemption, or donation while the vault is empty is therefore not covered. Could we call that out instead of saying this catches donation attacks through any entrypoint?

src/protection/vault/examples/MetaMorphoVaultAssertion.sol
```diff
@@ -59,7 +59,14 @@ contract MetaMorphoVaultAssertion is
...
     function triggers() external view override {
-        _registerSharePriceTriggers();
+        // Use the gas-bounded share-price triggers: MetaMorpho's computed `totalAssets()` loops the
+        // Morpho Blue supply queue, so the default envelope's all-fork-points `assetsMatchSharePrice`
+        // scan exhausts the assertion gas limit (PrecompileOOG) on unrelated high-call-count
+        // transactions that merely touch the vault, causing false invalidations. The bounded envelope
+        // compares only pre-tx vs post-tx (2 fork points) and is two-sided, so it still catches
+        // tx-wide dilution and donation/inflation through any entrypoint, while the per-call checks
+        // cover settlement during real deposit/mint/withdraw/redeem operations.
+        _registerBoundedSharePriceTriggers();
```
mateo-mro, line 69: Could we add a small PCL regression for this bundle? It can arm a mock ERC4626 through the tx end selector and run a transaction with several calls, which protects this new trigger wiring from a future regression. Replay validation confirms the reported window, but CI does not exercise this library path today.

mateo-mro (approved):

PR #90, opened 2026-07-21
feat(anomaly): anomaly score compositions

0xAlcibiades:


src/protection/anomaly/AnomalyGatedBaseAssertion.sol
```diff
@@ -0,0 +1,171 @@
...
+    /// @notice The watched contract whose anomaly score gates this assertion (the adopter).
+    address internal immutable target;
+
+    /// @notice Scores at or above this (out of 10_000) are treated as anomalous. On the Aave family
+    ///         the calibrated operating point for a 1% false-positive budget is 205, a 2% probability.
+    uint16 internal immutable anomalyThresholdBps;
+
+    /// @param _target The watched contract the model scores.
+    /// @param _anomalyThresholdBps The operating point, in bps of anomaly probability.
+    constructor(address _target, uint16 _anomalyThresholdBps) {
```
mateo-mro, line 87: Zero and values over 10,000 are valid `uint16` values here. At zero, the zero filled context returned for an unscored target satisfies the `>=` check, so a damage heuristic may revert without an anomaly score. Values over 10,000 make the gate unreachable. Please validate the range at deployment and cover its boundaries.

src/protection/anomaly/AnomalyGatedBaseAssertion.sol
```diff
@@ -0,0 +1,171 @@
...
+        registerAssertionSpec(AssertionSpec.Reshiram);
+        target = _target;
```
mateo-mro, line 95: The target has no zero address guard. `watchAnomaly` is registered for this value and `anomalyContext` returns zero when it cannot score the target, so a zero target creates an assertion that never evaluates its heuristics at a normal threshold. Treat it as a deployment error just like the other required addresses.

src/protection/anomaly/AnomalyCompositeAssertion.sol
```diff
@@ -0,0 +1,192 @@
...
+    bytes internal oracleQuery;
+
+    constructor(Config memory c) AnomalyGatedBaseAssertion(c.target, c.anomalyThresholdBps) {
+        if (!(c.bareGateBaseline || c.useDrain || c.useUpgrade || c.useAccounting || c.useOracle)) {
+            revert NoHeuristicEnabled();
+        }
+        // Each enabled leg must carry the parameters it reads; a zero address leaves the leg
+        // silently inert or falsely blocking, and a zero drain fraction corroborates on any
+        // transaction. The upgrade leg needs no check: a zero `upgradeTarget` means `target`.
+        if (c.useDrain && (c.outflowTarget == address(0) || c.outflowToken == address(0) || c.outflowFracBps == 0)) {
```
mateo-mro: `outflowFracBps` is only checked for zero. A normal ERC20 cannot lose more than its pretransaction balance, so any value above 10,000 leaves that enabled leg permanently silent. Please enforce the bound in `AnomalyGatedOutflowAssertion` too, otherwise the composite validates it while the reusable mixin stays inert.

src/protection/anomaly/AnomalyGatedBaseAssertion.sol
```diff
@@ -0,0 +1,188 @@
...
+        for (uint256 i; i < deltas.length; ++i) {
+            if (deltas[i].from == outflowTarget) {
+                outflow += deltas[i].value;
+            }
+            if (deltas[i].to == outflowTarget) {
+                inflow += deltas[i].value;
+            }
+        }
+        uint256 net = outflow > inflow ? outflow - inflow : 0;
+        return net * 10_000 / preBalance >= fracBps;
```
mateo-mro: The percentage calculation overflows before division when `net` is above `type(uint256).max / 10_000`. That turns an anomalous transaction with a below threshold drain into an arithmetic revert, which blocks without corroboration. `ph.mulDivDown` keeps the calculation in range.

mateo-mro (approved):

.github/workflows/solidity-test.yml
```diff
@@ -29,7 +29,12 @@ jobs:
           sudo install pcl/pcl /usr/local/bin/pcl
 
       - name: Run protection PCL tests
-        run: FOUNDRY_PROFILE=ci pcl test --match-path 'test/protection/**'
+        # pcl 1.5.1 runs suites on default-stack (2 MiB) worker threads and overflows once enough
```
odyslam, line 32: @0xAlcibiades if this is a bug on the pcl can you quickly push a fix there.

It's not clear to me if that's standard behavior from `pcl` or not.

0xAlcibiades: Got a sparate ticket open for it, idk if it's a bug or a need.

PR #93, opened 2026-07-24
feat(safe): add guard lifecycle scripts

makemake-kbo:
## Summary

- make CredibleSafeGuard fail open when registry reads revert, exhaust their bounded gas budget, or return malformed data
- add Foundry scripts to deploy the guard and generate checksummed Safe Transaction Builder JSON for installation and removal
- validate Safe/guard addresses and Safe guard ERC-165 support before generating a batch
- document the Safe{Wallet} import, review, signing, replacement, and removal workflow

## Testing

- `FOUNDRY_PROFILE=safe-guard forge test --offline -vv` — 20 passed
- `forge test --offline --match-path test/protection/safe/CredibleSafeGuard.t.sol -vv` — 29 passed
- `FOUNDRY_PROFILE=safe-guard forge build --offline`
- `forge fmt --check`
- `git diff --check`

### DIFF
--- a/examples/safe-guard/script/DeployCredibleSafeGuard.s.sol
+++ b/examples/safe-guard/script/DeployCredibleSafeGuard.s.sol
@@ -8,9 +8,22 @@ import {ICredibleRegistry} from "credible-std/protection/safe/ICredibleRegistry.
 
 /// @notice Deploys a Credible Safe guard using Foundry's configured broadcast wallet.
 contract DeployCredibleSafeGuard is Script {
+    /// @notice Thrown when the configured registry address has no deployed code.
+    /// @dev Catches the common footgun of a typo that resolves to an EOA: a codeless registry
+    ///      makes every credibility probe fail open, so the guard would allow every transaction.
+    error RegistryHasNoCode(address registry);
+    /// @notice Thrown when a required registry read (isCredibleBlock / lastCredibleBlock) does not
+    ///         return a single, well-formed 32-byte word.
+    error RegistryReadFailed(address registry, string read);
+
     function run() external returns (CredibleSafeGuard guard) {
         address registry = vm.envAddress("CREDIBLE_REGISTRY");
         uint256 threshold = vm.envUint("FAIL_OPEN_BLOCK_THRESHOLD");
+
+        // Validate the registry before broadcasting so a misconfigured (codeless or
+        // non-responsive) registry is caught up front rather than silently deploying a
+        // permanently-fail-open guard.
+        validateRegistry(registry);
 
         vm.startBroadcast();
         guard = deploy(registry, threshold);
@@ -24,5 +37,21 @@ contract DeployCredibleSafeGuard is Script {
 
     function deploy(address registry, uint256 threshold) public returns (CredibleSafeGuard) {
         return new CredibleSafeGuard(ICredibleRegistry(registry), threshold);
+    }
+
+    /// @notice Asserts the registry has code and answers both reads the guard depends on.
+    /// @dev Reverts with a descriptive error otherwise. Kept separate from {deploy} so the
+    ///      constructor's own zero-address / zero-threshold validation is preserved for callers
+    ///      that deploy against an in-memory mock.
+    function validateRegistry(address registry) public view {
+        if (registry.code.length == 0) revert RegistryHasNoCode(registry);
+
+        (bool credibleOk, bytes memory credibleData) =
+            registry.staticcall(abi.encodeCall(ICredibleRegistry.isCredibleBlock, (block.number)));
+        if (!credibleOk || credibleData.length != 32) revert RegistryReadFailed(registry, "isCredibleBlock");
+
+        (bool lastOk, bytes memory lastData) =
+            registry.staticcall(abi.encodeCall(ICredibleRegistry.lastCredibleBlock, ()));
+        if (!lastOk || lastData.length != 32) revert RegistryReadFailed(registry, "lastCredibleBlock");
     }
 }
--- a/examples/safe-guard/script/DeployCredibleSafeGuard.s.sol
+++ b/examples/safe-guard/script/DeployCredibleSafeGuard.s.sol
@@ -1,0 +1,28 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.13;
+
+import {Script, console2} from "forge-std/Script.sol";
+
+import {CredibleSafeGuard} from "credible-std/protection/safe/CredibleSafeGuard.sol";
+import {ICredibleRegistry} from "credible-std/protection/safe/ICredibleRegistry.sol";
+
+/// @notice Deploys a Credible Safe guard using Foundry's configured broadcast wallet.
+contract DeployCredibleSafeGuard is Script {
+    function run() external returns (CredibleSafeGuard guard) {
+        address registry = vm.envAddress("CREDIBLE_REGISTRY");
+        uint256 threshold = vm.envUint("FAIL_OPEN_BLOCK_THRESHOLD");
+
+        vm.startBroadcast();
+        guard = deploy(registry, threshold);
+        vm.stopBroadcast();
+
+        console2.log("Chain ID:", block.chainid);
+        console2.log("Credible Safe guard:", address(guard));
+        console2.log("Credible Registry:", registry);
+        console2.log("Fail-open block threshold:", threshold);
+    }
+
+    function deploy(address registry, uint256 threshold) public returns (CredibleSafeGuard) {
+        return new CredibleSafeGuard(ICredibleRegistry(registry), threshold);
+    }
+}
--- a/examples/safe-guard/script/DeployCredibleSafeGuard.s.sol
+++ b/examples/safe-guard/script/DeployCredibleSafeGuard.s.sol
@@ -46,9 +46,16 @@ contract DeployCredibleSafeGuard is Script {
     function validateRegistry(address registry) public view {
         if (registry.code.length == 0) revert RegistryHasNoCode(registry);
 
+        // `isCredibleBlock` is a Solidity `bool`, so a well-formed answer is a canonical boolean
+        // word (0 or 1). Reject any other 32-byte value (e.g. `abi.encode(uint256(2))`) here:
+        // it passes the length check but the guard's runtime decode treats it as unreadable
+        // (see CredibleSafeGuard._tryIsCredibleBlock's `value > 1` branch), which would otherwise
+        // let a registry silently deploy a permanently-fail-open guard.
         (bool credibleOk, bytes memory credibleData) =
             registry.staticcall(abi.encodeCall(ICredibleRegistry.isCredibleBlock, (block.number)));
-        if (!credibleOk || credibleData.length != 32) revert RegistryReadFailed(registry, "isCredibleBlock");
+        if (!credibleOk || credibleData.length != 32 || abi.decode(credibleData, (uint256)) > 1) {
+            revert RegistryReadFailed(registry, "isCredibleBlock");
+        }
 
         (bool lastOk, bytes memory lastData) =
             registry.staticcall(abi.encodeCall(ICredibleRegistry.lastCredibleBlock, ()));
--- a/examples/safe-guard/script/GenerateSafeGuardBatch.s.sol
+++ b/examples/safe-guard/script/GenerateSafeGuardBatch.s.sol
@@ -1,0 +1,35 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.13;
+
+import {console2} from "forge-std/Script.sol";
+
+import {SafeGuardBatch} from "./SafeGuardBatch.sol";
+
+/// @notice Generates a Safe Transaction Builder JSON file for installing or removing a guard.
+/// @dev This script never broadcasts or signs. Import the generated file into Safe{Wallet}'s
+///      Transaction Builder and submit it through the Safe's normal owner-confirmation flow.
+contract GenerateSafeGuardBatch is SafeGuardBatch {
+    function run() external returns (string memory json, string memory path) {
+        address safe = vm.envAddress("SAFE_ADDRESS");
+        Action action = parseAction(vm.envString("SAFE_GUARD_ACTION"));
+        address guard = vm.envOr("CREDIBLE_SAFE_GUARD", address(0));
+        uint256 createdAt = vm.unixTime();
+
+        if (action == Action.Install) {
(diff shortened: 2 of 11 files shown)


examples/safe-guard/script/GenerateSafeGuardBatch.s.sol
```diff
@@ -0,0 +1,35 @@
...
+pragma solidity ^0.8.13;
+
+import {console2} from "forge-std/Script.sol";
+
+import {SafeGuardBatch} from "./SafeGuardBatch.sol";
+
+/// @notice Generates a Safe Transaction Builder JSON file for installing or removing a guard.
+/// @dev This script never broadcasts or signs. Import the generated file into Safe{Wallet}'s
+///      Transaction Builder and submit it through the Safe's normal owner-confirmation flow.
+contract GenerateSafeGuardBatch is SafeGuardBatch {
```
mateo-mro, line 11: The Linear issue asks for installation and removal scripts that work end to end against Anvil. This script only writes an unsigned Transaction Builder file, so neither action changes the Safe on the local node. Can we add an Anvil execution path that submits the owner authorized Safe transaction, or align the issue acceptance criteria with this manual workflow?

src/protection/safe/CredibleSafeGuard.sol
```diff
@@ -132,31 +138,69 @@ contract CredibleSafeGuard is ITransactionGuard {
...
-    ///      offline. The credible-block check runs first so the expected hot path (credible block,
-    ///      live builder set) costs a single registry call.
+    /// @dev Core gate: allow credible blocks, otherwise fail open when a required registry read
+    ///      fails or the builder set is offline. The credibility check runs first so the expected
+    ///      hot path (credible block, live builder set) costs a single registry call.
     function _checkCredibleBlock() internal view {
-        if (credibleRegistry.isCredibleBlock(block.number)) return;
-        if (!_failOpenActive()) revert NonCredibleBlock();
+        (bool readable, bool credible) = _tryIsCredibleBlock(block.number);
+        if (!readable || credible || _failOpenActive()) return;
```
mateo-mro, line 194: This changes the production guard from reverting when a registry call fails to allowing every owner transaction, while the Linear issue is scoped to lifecycle scripts. That is a material change to the enforcement boundary during registry outages. Can we move it into a separately approved change, or link the product requirement that authorizes this policy?

examples/safe-guard/script/DeployCredibleSafeGuard.s.sol
```diff
@@ -0,0 +1,28 @@
...
+import {ICredibleRegistry} from "credible-std/protection/safe/ICredibleRegistry.sol";
+
+/// @notice Deploys a Credible Safe guard using Foundry's configured broadcast wallet.
+contract DeployCredibleSafeGuard is Script {
+    function run() external returns (CredibleSafeGuard guard) {
+        address registry = vm.envAddress("CREDIBLE_REGISTRY");
+        uint256 threshold = vm.envUint("FAIL_OPEN_BLOCK_THRESHOLD");
+
+        vm.startBroadcast();
+        guard = deploy(registry, threshold);
```
mateo-mro, line 29: Can the deployment path reject a registry with no code and verify both required reads before broadcasting? A typo that resolves to an EOA currently deploys successfully, and the resulting guard allows every transaction because each registry probe activates fail open. The later batch validation only checks the guard interface, so it cannot catch this configuration error.

src/protection/safe/CredibleSafeGuard.sol
```diff
@@ -78,6 +80,10 @@ interface ITransactionGuard is IERC165 {
 ///      Both the registry address and the fail-open threshold are immutable; re-pointing or
 ///      re-tuning means deploying a new guard and calling `setGuard` again.
 contract CredibleSafeGuard is ITransactionGuard {
+    /// @dev Bounds the gas and returndata exposure of each registry probe. A registry that cannot
+    ///      answer within this budget is treated as unavailable and the guard fails open.
+    uint256 internal constant REGISTRY_READ_GAS_LIMIT = 50_000;
```
mateo-mro, line 109: Can we add a registry test double that consumes more than the 50000 gas budget and another that returns more than 32 bytes? Those are the adversarial cases this bounded call exists to handle, but the tests currently cover only reverts, short data, a noncanonical boolean, and no code. This would prove the call fails open without copying unbounded return data.

examples/safe-guard/script/DeployCredibleSafeGuard.s.sol
```diff
@@ -0,0 +1,57 @@
...
+    /// @notice Asserts the registry has code and answers both reads the guard depends on.
+    /// @dev Reverts with a descriptive error otherwise. Kept separate from {deploy} so the
+    ///      constructor's own zero-address / zero-threshold validation is preserved for callers
+    ///      that deploy against an in-memory mock.
+    function validateRegistry(address registry) public view {
+        if (registry.code.length == 0) revert RegistryHasNoCode(registry);
+
+        (bool credibleOk, bytes memory credibleData) =
+            registry.staticcall(abi.encodeCall(ICredibleRegistry.isCredibleBlock, (block.number)));
+        if (!credibleOk || credibleData.length != 32) revert RegistryReadFailed(registry, "isCredibleBlock");
```
mateo-mro: `isCredibleBlock` is a Solidity `bool`, but this accepts any 32 byte word. A registry returning `2` passes validation, while `_tryIsCredibleBlock` rejects the same value at runtime and the deployed guard then allows every owner transaction. I reproduced this with a regression test that expects `validateRegistry` to reject `abi.encode(uint256(2))`; it currently does not revert. Can we decode the word as a canonical bool here before broadcasting?

src/protection/safe/CredibleSafeGuard.sol
```diff
@@ -132,31 +160,85 @@ contract CredibleSafeGuard is ITransactionGuard {
...
-    ///      reports a last credible block at or beyond the current block.
+    /// @dev Fail-open is active when the last-block read fails or the current block is strictly
+    ///      more than `failOpenBlockThreshold` blocks ahead of the last credible block. The
+    ///      `block.number > lastCredibleBlock_` guard avoids underflow if the registry reports a
+    ///      last credible block at or beyond the current block.
     function _failOpenActive() internal view returns (bool) {
-        uint256 lastCredibleBlock_ = credibleRegistry.lastCredibleBlock();
+        (bool readable, uint256 lastCredibleBlock_) = _tryLastCredibleBlock();
+        if (!readable) return true;
         return block.number > lastCredibleBlock_ && block.number - lastCredibleBlock_ > failOpenBlockThreshold;
```
mateo-mro: The documented recovery guarantee is not true for a future `lastCredibleBlock`. If a broken registry returns `type(uint256).max` and false for the current block, this branch keeps fail open disabled and every owner transaction reverts until the reported height is reached. That is effectively permanent, and the current test explicitly preserves it. Since the registry defines this value as the highest block marked so far, a value above `block.number` is invalid and should be treated as an unreadable response if the goal is to prevent registry failures from bricking the Safe.

makemake-kbo: Fixed in 22434a1. `_failOpenActive` now treats `lastCredibleBlock_ > block.number` as an unreadable/broken response and fails open, so a future height (including `type(uint256).max`) can no longer keep fail-open disabled and brick the Safe. The old `test_noUnderflow_whenLastCredibleBlockInFuture` (which preserved the revert) is replaced by `test_failsOpen_whenLastCredibleBlockInFuture` and `test_failsOpen_whenLastCredibleBlockIsMaxUint`, plus a real-Safe E2E `test_realSafe_failsOpenWhenLastBlockReportedInFuture`. `test_reverts_whenLastCredibleBlockEqualsCurrent_andNotCredible` pins that an equal (gap-0) reading stays a normal blocked tx.

src/protection/safe/CredibleSafeGuard.sol
makemake-kbo: Both adversarial doubles are covered in `test/protection/safe/CredibleSafeGuard.t.sol`: `GasGuzzlingRegistry` burns >> 50k gas on either read (toggled via `guzzleIsCredible`) and drives `test_failsOpen_whenCredibilityReadExceedsGasBudget` / `test_failsOpen_whenLastCredibleBlockReadExceedsGasBudget`; `ReturnDataBombRegistry` returns 3200 bytes with a leading truthy word and drives `test_failsOpen_whenRegistryReturnsMoreThan32Bytes`. All three assert the guard fails open without copying unbounded returndata (retSize is capped at one word + a `returndatasize() == 0x20` check).

examples/safe-guard/script/DeployCredibleSafeGuard.s.sol
makemake-kbo: Addressed in e23f54d: `validateRegistry` now runs before `vm.startBroadcast`, rejects a codeless/EOA registry (`RegistryHasNoCode`), and verifies both `isCredibleBlock` and `lastCredibleBlock` return a well-formed 32-byte word (`RegistryReadFailed`) before deploying. Covered by `test_validateRegistry_rejectsCodelessRegistry` and `test_validateRegistry_rejectsUnresponsiveRegistry`.

examples/safe-guard/script/DeployCredibleSafeGuard.s.sol
makemake-kbo: Fixed in bb3d6c5: `validateRegistry` now decodes the `isCredibleBlock` word as a canonical bool and rejects any value > 1 (e.g. `abi.encode(uint256(2))`) before broadcasting, matching the guards `_tryIsCredibleBlock` runtime decode. Regression test `test_validateRegistry_rejectsNonCanonicalBoolRegistry`.

examples/safe-guard/script/GenerateSafeGuardBatch.s.sol
makemake-kbo: The install/remove flow is intentionally the unsigned Transaction Builder JSON + Safe{Wallet} path rather than a broadcasting script: a production guard target is a multisig Safe, so a single script signer cannot legitimately submit an owner-authorized `execTransaction` on its behalf, and doing so from a script would only work for a degenerate 1-of-1 Safe.

The end-to-end behavior the criteria are after is covered by the integration tests, which install and remove the guard through a real Gnosis Safe (v1.4.1) via owner-signed `execTransaction`: `test_generatedInstallBatch_executesThroughRealSafe` and `test_generatedBatch_replacesAndRemovesGuard` in `CredibleSafeGuardScripts.t.sol`, and the full install/allow/block/fail-open lifecycle in `CredibleSafeGuardSafeIntegration.t.sol`. These exercise the exact `setGuard` payload the generator emits.

Proposal: align the ENG-4054 acceptance criteria to "generate an importable, checksum-valid Transaction Builder batch for install/remove, with E2E coverage against a real Safe in tests" rather than a live-Anvil broadcast script. Happy to adjust if you would rather ship a broadcast helper scoped to a 1-of-1 test Safe.

src/protection/safe/CredibleSafeGuard.sol
makemake-kbo: Agreed this is a real enforcement-boundary decision and not an accidental byproduct of the bounded-call hardening. The intent is: this guard must never permanently brick the Safe owner path because of an unavailable/broken/misconfigured registry, so registry failure degrades to "no credibility protection" rather than a frozen Safe. That is now written up explicitly in the contract-level NatSpec (05213a9) including the stated security tradeoff.

The tradeoff (a broken/absent/malicious registry lets all owner txs through) is mitigated at deploy time, not runtime: `validateRegistry` rejects a codeless/EOA registry and verifies both reads return well-formed, canonical data before broadcasting (e23f54d, bb3d6c5), and the registry address is immutable, so a guard is never deployed already pointed at a permanently-fail-open registry. 22434a1 additionally closes the one case where a broken registry could brick rather than fail open (future `lastCredibleBlock`).

Since it is intrinsic to how the guard behaves during a registry outage, I would prefer to keep it in this PR rather than split it. Can you approve the fail-open policy inline here, or point me at a product requirement to cite? Happy to carve it into its own change if you would rather review the policy separately from the scripts.

examples/safe-guard/script/DeployCredibleSafeGuard.s.sol
```diff
@@ -0,0 +1,64 @@
...
+    function validateRegistry(address registry) public view {
+        if (registry.code.length == 0) revert RegistryHasNoCode(registry);
+
+        // `isCredibleBlock` is a Solidity `bool`, so a well-formed answer is a canonical boolean
+        // word (0 or 1). Reject any other 32-byte value (e.g. `abi.encode(uint256(2))`) here:
+        // it passes the length check but the guard's runtime decode treats it as unreadable
+        // (see CredibleSafeGuard._tryIsCredibleBlock's `value > 1` branch), which would otherwise
+        // let a registry silently deploy a permanently-fail-open guard.
+        (bool credibleOk, bytes memory credibleData) =
+            registry.staticcall(abi.encodeCall(ICredibleRegistry.isCredibleBlock, (block.number)));
```
mateo-mro, line 55: `validateRegistry` uses an unrestricted `staticcall`, while the deployed guard caps the same read at 50000 gas. A registry that returns a valid boolean after more than that budget passes this check, but every runtime probe is then unreadable and the guard allows all owner transactions. I reproduced this with a registry whose boolean read performs a bounded hashing loop: validation succeeds, while the deployed guard immediately reports fail open. Can this preflight use the same gas and return data bounds as `_tryIsCredibleBlock` and `_tryLastCredibleBlock`?

examples/safe-guard/script/DeployCredibleSafeGuard.s.sol
```diff
@@ -0,0 +1,64 @@
...
+        if (!credibleOk || credibleData.length != 32 || abi.decode(credibleData, (uint256)) > 1) {
+            revert RegistryReadFailed(registry, "isCredibleBlock");
+        }
+
+        (bool lastOk, bytes memory lastData) =
+            registry.staticcall(abi.encodeCall(ICredibleRegistry.lastCredibleBlock, ()));
+        if (!lastOk || lastData.length != 32) revert RegistryReadFailed(registry, "lastCredibleBlock");
```
mateo-mro, line 62: `validateRegistry` also accepts any 32 byte `lastCredibleBlock`, including `type(uint256).max`. After 22434a1 the deployed guard treats any future value as a broken read and immediately fails open, so this can still deploy the exact unsafe configuration the preflight is meant to reject. I reproduced it with a registry returning `false` and `type(uint256).max`; validation did not revert. Can we reject values above `block.number` here as well?

mateo-mro (approved):

PR #94, opened 2026-07-24
test(credible-block): live-anvil scripts for credible block guard upgrades

makemake-kbo:
## What

Adds an `examples/credible-block` project with scripts to test Credible Layer contract upgrades against a **live anvil node**.

The existing forge unit tests fake block state with `vm.roll`/`vm.prank`. That covers the guard's decision logic but cannot reproduce the one thing that only exists on a real chain: a builder's *credible-block marker* transaction and a *guarded* transaction landing in the **same block** (a bundle). These scripts seed a real node and drive mining manually to test exactly that.

## Contents

- **`src/CredibleRegistry.sol`** — minimal `ICredibleRegistry`: a single immutable builder (set at construction) can mark the current block credible. Trimmed to the least required to exercise the guard; the production registry additionally has a timelocked admin, builder whitelist, and timestamp slot-binding.
- **`src/GuardedCounter.sol`** — a concrete `CredibleBlockGuard` consumer standing in for an upgraded credible-layer contract (`bump()` is `onlyCredibleBlock`).
- **`script/test-credible-upgrades.sh`** — boots anvil (fifo mempool ordering), deploys the registry + a builder we control + the guard, then verifies three cases.
- **`foundry.toml`** — new `[profile.credible-block]`.

## Cases verified (all pass)

1. **Credible block** — bundle `[markCurrentBlockCredible, bump]` into one block (automine off + `cast send --async` + single `evm_mine`); both succeed, counter increments.
2. **Non-credible block** — `bump()` with no marker reverts with `NonCredibleBlock()`; counter unchanged.
3. **Fail-open** — after the builder stays silent for `> failOpenBlockThreshold` blocks, `bump()` passes again (and still reverts at the boundary, gap == threshold).

## Running

```shell
./examples/credible-block/script/test-credible-upgrades.sh
```

Requires `anvil`, `cast`, `forge`, `jq`. Exits non-zero on any failed check. See `examples/credible-block/README.md` for details.

### DIFF
--- a/examples/credible-block/README.md
+++ b/examples/credible-block/README.md
@@ -15,7 +15,7 @@ a real node and drive mining manually so we can test exactly that.
 
 | File | Role |
 | ---- | ---- |
-| [`src/CredibleRegistry.sol`](./src/CredibleRegistry.sol) | Minimal deployable registry: a single immutable builder (set at construction) can mark the current block credible; implements [`ICredibleRegistry`](../../src/protection/credible_block/ICredibleRegistry.sol). The production registry ([`phylaxsystems/credible-registry`](../../../credible-registry)) additionally has a timelocked admin, a builder whitelist, and timestamp slot-binding — none needed to exercise the guard. |
+| [`src/CredibleRegistry.sol`](./src/CredibleRegistry.sol) | Minimal deployable registry: a single immutable builder (set at construction) can mark the current block credible; implements [`ICredibleRegistry`](../../src/protection/credible_block/ICredibleRegistry.sol). The production registry ([`phylaxsystems/credible-registry`](https://github.com/phylaxsystems/credible-registry)) additionally has a timelocked admin, a builder whitelist, and timestamp slot-binding — none needed to exercise the guard. |
 | [`src/GuardedCounter.sol`](./src/GuardedCounter.sol) | A concrete `CredibleBlockGuard` standing in for an upgraded credible-layer contract; its `bump()` entrypoint is `onlyCredibleBlock`. |
 | [`script/test-credible-upgrades.sh`](./script/test-credible-upgrades.sh) | Orchestrator that boots anvil, deploys, and runs the three cases. |
 
--- a/examples/credible-block/README.md
+++ b/examples/credible-block/README.md
@@ -1,0 +1,55 @@
+# Credible Block guard — upgrade tests
+
+Integration scripts that exercise [`CredibleBlockGuard`](../../src/protection/credible_block/CredibleBlockGuard.sol)'s
+`onlyCredibleBlock` modifier against a **live anvil node**, so we can validate credible-layer
+contract upgrades end to end.
+
+The forge unit tests in
+[`test/protection/credible_block/`](../../test/protection/credible_block/CredibleBlockGuard.t.sol)
+fake block state with `vm.roll` / `vm.prank`. That covers the pure decision logic, but it cannot
+reproduce the one thing that only exists on a real chain: a builder's *credible block marker*
+transaction and a *guarded* transaction landing in the **same block** (a bundle). These scripts seed
+a real node and drive mining manually so we can test exactly that.
+
+## Contents
+
+| File | Role |
+| ---- | ---- |
+| [`src/CredibleRegistry.sol`](./src/CredibleRegistry.sol) | Minimal deployable registry: a single immutable builder (set at construction) can mark the current block credible; implements [`ICredibleRegistry`](../../src/protection/credible_block/ICredibleRegistry.sol). The production registry ([`phylaxsystems/credible-registry`](../../../credible-registry)) additionally has a timelocked admin, a builder whitelist, and timestamp slot-binding — none needed to exercise the guard. |
+| [`src/GuardedCounter.sol`](./src/GuardedCounter.sol) | A concrete `CredibleBlockGuard` standing in for an upgraded credible-layer contract; its `bump()` entrypoint is `onlyCredibleBlock`. |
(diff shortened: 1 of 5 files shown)


examples/credible-block/script/test-credible-upgrades.sh
```diff
@@ -0,0 +1,196 @@
...
+# --------------------------------------------------------------------------------------------------
+section "Booting anvil (fifo mempool ordering) and deploying contracts"
+
+# fifo ordering guarantees the marker tx (submitted first) is included before the guarded tx when
+# both share a block, so the guarded call sees the block already marked credible.
+anvil --port "$RPC_PORT" --order fifo --silent &
+ANVIL_PID=$!
+
+for _ in $(seq 1 50); do
+    cast block-number --rpc-url "$RPC_URL" >/dev/null 2>&1 && break
```
mateo-mro, line 113: If the selected port already has an RPC listener, the new Anvil process exits but this loop accepts that other listener as ready. The script then snapshots, reverts, and disables automine on the existing node. Verify the child process is still alive before continuing, or require an unused port.

examples/credible-block/script/test-credible-upgrades.sh
```diff
@@ -0,0 +1,214 @@
...
+# Requires: anvil, cast, forge, jq on PATH.
+
+set -euo pipefail
+
+# --------------------------------------------------------------------------------------------------
+# Config
+# --------------------------------------------------------------------------------------------------
+RPC_PORT="${RPC_PORT:-8545}"
+RPC_URL="http://127.0.0.1:${RPC_PORT}"
+FAIL_OPEN_THRESHOLD="${FAIL_OPEN_THRESHOLD:-10}" # kept small so the fail-open case runs quickly
```
mateo-mro, line 32: The documented `FAIL_OPEN_THRESHOLD` override can make Case 2 report a false failure for values the guard accepts. With a threshold of 1 or 2, deployment has already advanced the chain far enough that `lastCredibleBlock == 0` activates fail open before the noncredible transaction is mined. Please either validate the minimum required by this fixture or seed a credible block before testing the noncredible path.

examples/credible-block/script/test-credible-upgrades.sh
```diff
@@ -0,0 +1,214 @@
+#!/usr/bin/env bash
+#
+# Integration test for the Credible Layer block guard against a live anvil node.
+#
+# Unlike the forge unit tests (which fake block state with vm.roll/vm.prank), this drives a real
+# node so we can exercise the one thing that only matters on a real chain: whether a builder's
+# "credible block marker" tx and a guarded tx land in the SAME block (a bundle).
+#
+# It seeds anvil with:
+#   - a CredibleRegistry (examples/credible-block/src/CredibleRegistry.sol) whose admin is an EOA we
```
mateo-mro: This setup description is stale. `CredibleRegistry` has no admin or whitelist; it receives one immutable builder directly in the constructor. Please update these lines so the script does not claim to exercise behavior that is absent from the fixture.

examples/credible-block/README.md
```diff
@@ -0,0 +1,55 @@
...
+fake block state with `vm.roll` / `vm.prank`. That covers the pure decision logic, but it cannot
+reproduce the one thing that only exists on a real chain: a builder's *credible block marker*
+transaction and a *guarded* transaction landing in the **same block** (a bundle). These scripts seed
+a real node and drive mining manually so we can test exactly that.
+
+## Contents
+
+| File | Role |
+| ---- | ---- |
+| [`src/CredibleRegistry.sol`](./src/CredibleRegistry.sol) | Minimal deployable registry: a single immutable builder (set at construction) can mark the current block credible; implements [`ICredibleRegistry`](../../src/protection/credible_block/ICredibleRegistry.sol). The production registry ([`phylaxsystems/credible-registry`](../../../credible-registry)) additionally has a timelocked admin, a builder whitelist, and timestamp slot-binding — none needed to exercise the guard. |
```
mateo-mro: This relative target stays inside the current repository path, so it does not reach the sibling production registry repository when the README is rendered. Please use its absolute GitHub URL.

examples/credible-block/script/test-credible-upgrades.sh
makemake-kbo: Fixed in 380b169: the script now refuses to run if the port already has an RPC listener, and the readiness loop checks the anvil child is still alive (bailing if it exited) before proceeding, so it can no longer adopt a foreign node.

examples/credible-block/script/test-credible-upgrades.sh
makemake-kbo: Fixed in 6afb544: Case 2 now seeds a credible block first, so the guarded tx lands at gap 1. Since the guard rejects a zero threshold, every deployable threshold is >= 1 and gap 1 is never > threshold, so fail-open stays inactive for any FAIL_OPEN_THRESHOLD override.

examples/credible-block/script/test-credible-upgrades.sh
makemake-kbo: Fixed in 6afb544: the header comment now describes the single immutable builder set at construction and notes the timelocked admin / whitelist live only in the production registry and aren't needed here.

examples/credible-block/README.md
makemake-kbo: Fixed in 6afb544: the README now links the production registry by its absolute GitHub URL (https://github.com/phylaxsystems/credible-registry).

mateo-mro (approved):

PR #95, opened 2026-07-24
feat: initial protocol manager interface + CredibleSafeGuard integration

makemake-kbo:
## What

Adds the **initial protocol manager** primitive to `credible-std` and wires it into `CredibleSafeGuard`.

A protected contract exposes `initialProtocolManager()` to declare the address allowed to manage its assertions in the Credible Layer. The state oracle reads it when the protocol is initialized, so **deploying the contract is the ownership proof** — no separate manual review/claim round.

## Changes

**New primitive** (`src/protection/initial_protocol_manager/`)
- `IInitialProtocolManager` — the read interface the state oracle calls.
- `InitialProtocolManager` — abstract base fixing the manager immutably at deployment, with a zero-address guard.
- README documenting the concept and the inherit-and-forward usage.

**CredibleSafeGuard integration** (`src/protection/safe/CredibleSafeGuard.sol`)
- Now `is ITransactionGuard, InitialProtocolManager`; constructor takes the manager address and forwards it to the base.
- Deploy script + example wiring read a new `INITIAL_PROTOCOL_MANAGER` env var.

**Tests**
- `InitialProtocolManager.t.sol` — construction, fuzz over non-zero managers, and interface conformance (readable through the `IInitialProtocolManager` type).
- `CredibleSafeGuard.t.sol` — asserts the getter, zero-manager revert, and interface conformance.
- Updated example suites for the new constructor arg.

## Verification
- `forge build`: compiles (pre-existing lint notes only).
- Safe guard suite: 31 passed. Initial-protocol-manager suite: 4 passed. Example `safe-guard` suites: 20 passed.


src/protection/credible_block/IInitialProtocolManager.sol
```diff
@@ -0,0 +1,24 @@
...
+/// @dev The state oracle calls {initialProtocolManager} on the contract when the protocol is
+///      initialized in the Credible Layer and registers the returned address as the manager.
+///      Because the value is defined by the contract's own code, deploying the contract is the
+///      ownership proof: whoever controlled the deployment chose the manager. This is what makes
+///      updated or redeployed contracts self-verifying, with no separate claim step.
+///
+///      Implementing this interface is optional. Contracts that do not expose it (for example,
+///      already-deployed contracts that cannot be changed) go through manual verification, where
+///      Phylax confirms ownership directly and sets the manager.
+interface IInitialProtocolManager {
```
mateo-mro, line 19: ENG 4072 defines the public import as `credible-std/protection/credible_block/IInitialProtocolManager.sol` and asks for the example in that module README. This different module path means an integrator following the acceptance criteria cannot import it, and the requested documentation remains unchanged. Please put the interface and example at the promised path.

src/protection/safe/CredibleSafeGuard.sol
```diff
@@ -77,7 +80,16 @@ interface ITransactionGuard is IERC165 {
 ///        - ~1s blocks:                     15 min  ~= 900 blocks
 ///      Both the registry address and the fail-open threshold are immutable; re-pointing or
 ///      re-tuning means deploying a new guard and calling `setGuard` again.
-contract CredibleSafeGuard is ITransactionGuard {
+///
+///      Onboarding. The guard also implements {IInitialProtocolManager} (via {InitialProtocolManager}),
+///      exposing the address allowed to manage this deployment's assertions in the Credible Layer.
+///      The state oracle reads it when the protocol is initialized, so deploying the guard is itself
+///      the ownership proof — no separate manual review round.
+contract CredibleSafeGuard is ITransactionGuard, InitialProtocolManager {
```
mateo-mro, line 112: ENG 4072 explicitly requires a standalone interface with no guard coupling. Inheriting the new base here also changes the deployed guard ABI by adding a required constructor argument, so existing integrations can no longer construct it unchanged. Please keep this PR to the interface and documentation and move guard adoption to a separate task.

examples/safe-guard/script/DeployCredibleSafeGuard.s.sol
```diff
@@ -0,0 +1,33 @@
+// SPDX-License-Identifier: MIT
```
mateo-mro, line 1: This guard lifecycle work belongs to ENG 4054 and is already being reviewed in PR 93. Carrying it here turns ENG 4072 into a 15 file behavioral change and duplicates code that has advanced on another branch. Please rebase this PR onto a clean base and keep only the standalone interface, requested README update, and optional test.

src/protection/safe/CredibleSafeGuard.sol
```diff
@@ -132,31 +148,69 @@ contract CredibleSafeGuard is ITransactionGuard {
...
-    ///      reports a last credible block at or beyond the current block.
+    /// @dev Fail-open is active when the last-block read fails or the current block is strictly
+    ///      more than `failOpenBlockThreshold` blocks ahead of the last credible block. The
+    ///      `block.number > lastCredibleBlock_` guard avoids underflow if the registry reports a
+    ///      last credible block at or beyond the current block.
     function _failOpenActive() internal view returns (bool) {
-        uint256 lastCredibleBlock_ = credibleRegistry.lastCredibleBlock();
+        (bool readable, uint256 lastCredibleBlock_) = _tryLastCredibleBlock();
+        if (!readable) return true;
         return block.number > lastCredibleBlock_ && block.number - lastCredibleBlock_ > failOpenBlockThreshold;
```
mateo-mro: When the registry returns a well formed block above `block.number`, this leaves fail open disabled. If `isCredibleBlock` is false, every owner transaction then reverts until that height, and `type(uint256).max` makes the lock effectively permanent. Since this change promises registry failures cannot brick the Safe, a future value should be treated as an unreadable response and covered with a maximum value regression test.

examples/safe-guard/script/DeployCredibleSafeGuard.s.sol
```diff
@@ -0,0 +1,33 @@
...
+import {ICredibleRegistry} from "credible-std/protection/safe/ICredibleRegistry.sol";
+
+/// @notice Deploys a Credible Safe guard using Foundry's configured broadcast wallet.
+contract DeployCredibleSafeGuard is Script {
+    function run() external returns (CredibleSafeGuard guard) {
+        address registry = vm.envAddress("CREDIBLE_REGISTRY");
+        uint256 threshold = vm.envUint("FAIL_OPEN_BLOCK_THRESHOLD");
+        address protocolManager = vm.envAddress("INITIAL_PROTOCOL_MANAGER");
+
+        vm.startBroadcast();
```
mateo-mro, line 29: Before broadcasting, please validate that the registry has code and that both reads satisfy the same gas and return value rules as the guard. A typo that resolves to an EOA currently deploys a guard whose probes are always unreadable, so every owner transaction is allowed. PR 93 already contains a newer validation path that should not be forked here.

lean-apple (approved):

PR #97, opened 2026-07-27
feat(anomaly): model preview roc curve

0xAlcibiades:


src/protection/anomaly/AnomalyGatedBaseAssertion.sol
```diff
@@ -77,43 +83,42 @@ abstract contract AnomalyGatedBaseAssertion is Assertion {
...
+    /// @notice Register the anomaly trigger for `selector` at this assertion's sensitivity level.
+    /// @dev The trigger is the gate. `selector` runs only when the AnomalySubsystem scores `target`
+    ///      anomalously enough to clear `sensitivity`, so the assertion body checks damage and
+    ///      nothing else. A transaction below the level never spawns the assertion at all.
+    ///
+    ///      The gate fails open at both ends. An unscored target, or one whose model carries no
+    ///      resolved ladder, clears no level, so nothing fires. Call this inside your `triggers()`.
+    ///
+    ///      `ph.anomalyContext(target)` still reports the score and the level it cleared, for an
+    ///      assertion that wants to act on how anomalous a transaction was.
```
fredo: `AnomalyContext` no longer exposes the raw score after `scoreBps` was replaced with `firesAt`; it reports only the highest sensitivity level cleared. Please describe only that level/verdict here, and update the equivalent wording in `TriggerRecorder.sol`, so adopters are not directed to a value they cannot read.

test/backtesting/UsdcOpSepoliaBacktesting.t.sol
```diff
@@ -57,36 +57,34 @@ contract BacktestingIntegrationTest is CredibleTestWithBacktesting {
...
 
-            BacktestingTypes.BacktestingResults memory results = executeBacktestForTransaction(
-                txHash,
-                USDC_OP_SEPOLIA,
-                type(ERC20Assertion).creationCode,
-                ERC20Assertion.assertionTransferInvariant.selector,
-                rpcUrl
-            );
+        assertEq(results.totalTransactions, 1, "the pinned transaction is the only one backtested");
+        assertEq(results.assertionFailures, 0, "a plain transfer keeps the invariant");
```
fredo, line 88: This does not prove that the assertion executed successfully. `BacktestingResults` tracks `successfulValidations`, `skippedTransactions`, `replayFailures`, and `unknownErrors` independently, so a skipped, replay-failed, or errored transaction can still satisfy `totalTransactions == 1` and `assertionFailures == 0`. Please assert `successfulValidations == 1` and that the other non-success counters are zero (or use a helper that enforces exactly one successful validation).

mateo-mro (dismissed):

makemake-kbo (dismissed):

test/protection/anomaly/AnomalyUngatedAssertion.t.sol
```diff
@@ -0,0 +1,57 @@
...
+    function setUp() public {
+        token = new MockERC20();
+        vault = new Vault(token);
+    }
+
+    /// One invalidation per firing, with nothing left to decide in the body.
+    function test_body_reverts_unconditionally() public {
+        UngatedHarness bare = new UngatedHarness(address(vault), Sensitivity.RECOMMENDED);
+
+        vm.expectRevert();
```
fredo: nit: Could we explicitly expect `AnomalousTransaction` here, ideally including the expected `firesAt`, instead of accepting any revert? We have encountered this class of false positive in the past: a broad `expectRevert()` can pass on an earlier, unrelated failure. Here `ph.anomalyContext(target)` can revert before the custom error is reached under a runtime without the precompile, so the test would pass without proving the intended behavior. Mock the context response if needed, then assert the encoded custom error.

src/protection/anomaly/AnomalyCompositeAssertion.sol
```diff
@@ -134,10 +134,6 @@ contract AnomalyCompositeAssertion is AnomalyGatedBaseAssertion {
     ///      fold hits the operator's absorbing value: a silent leg under AND, a corroborating leg
     ///      under OR. The alert cell (`a AND NOT H`) is the deliberate fall-through with no revert.
     function assertComposite() external {
```
mateo-mro, line 138: The disposition above still says `NOT a` returns early. With this gate removed, the trigger prevents this function from being dispatched at all. Could we update that sentence so the composite docs match the new execution path and cost?

test/protection/anomaly/AnomalyCompositeAssertion.t.sol
```diff
@@ -3,54 +3,46 @@ pragma solidity ^0.8.28;
...
 }
 
 /// @notice The composite with a protocol-specific `_extra` leg. The leg corroborates when the
 /// protocol reports itself unhealthy (`flag == false`) post-tx.
 contract CompositeWithHealthTxEnd is AnomalyCompositeAssertion {
-    bool internal immutable anomalous;
     address internal immutable healthTarget;
 
-    constructor(Config memory c, address healthTarget_, bool anomalous_) AnomalyCompositeAssertion(c) {
+    constructor(Config memory c, address healthTarget_) AnomalyCompositeAssertion(c) {
```
mateo-mro, line 36: This constructor now takes only the config and health target, but `_register` still appends the removed boolean at line 312. Solidity silently accepts those trailing constructor bytes, so the tests pass despite the stale encoding. Please remove the final `true` there.

test/protection/anomaly/AnomalyGatedBaseAssertion.t.sol
```diff
@@ -34,18 +35,18 @@ contract TestAnomalyDrainRatio is CredibleTest, Test {
...
         c.outflowFracBps = fracBps;
     }
 
-    function _register(uint256 fracBps, bool anomalous) internal {
+    function _register(uint256 fracBps) internal {
         cl.assertion({
             adopter: address(vault),
             createData: abi.encodePacked(
-                type(CompositeTxEndHarness).creationCode, abi.encode(_config(address(token), fracBps), anomalous)
+                type(CompositeTxEndHarness).creationCode, abi.encode(_config(address(token), fracBps))
```
mateo-mro, line 49: The codeless registration at line 111 still appends the removed anomaly boolean to this harness constructor. Solidity ignores the trailing bytes, which lets the test pass with a stale deployment encoding. Please remove the final `true` there.

makemake-kbo (approved):
Reviewed the sensitivity value type, trigger registration, anomaly assertion behavior, generated interfaces, and tests. The implementation matches the new verdict-based execution seam and the required checks are green.

mateo-mro (approved):

PR #98, opened 2026-07-30
Make testing guard upgrades more ergonomic

makemake-kbo:
## Summary

- generalize the live-Anvil credible-block integration runner to test a caller-specified upgraded contract and registry configuration
- verify guard configuration, expected revert behavior, strict fail-open boundaries, and caller-provided state effects
- keep `GuardedCounter` as a clearly labeled default fixture and document real-upgrade usage
- add argument parsing and failure-message regression tests

## Validation

- `./examples/credible-block/test/test-script-arguments.sh` — 7 passed
- default live-Anvil fixture run — 14 passed, 0 failed
- configurable deployment-command live-Anvil run — 14 passed, 0 failed
- `bash -n` and `git diff --check` — passed

### DIFF
--- a/examples/credible-block/script/test-credible-upgrades.sh
+++ b/examples/credible-block/script/test-credible-upgrades.sh
@@ -19,8 +19,11 @@ TARGET_DEPLOY_COMMAND=""
 REGISTRY_ADDRESS=""
 REGISTRY_DEPLOY_COMMAND=""
 GUARDED_CALL=""
+GUARDED_CALL_ARGS=()
 MARKER_CALL="markCurrentBlockCredible()"
+MARKER_CALL_ARGS=()
 STATE_READ_CALL=""
+STATE_READ_CALL_ARGS=()
 STATE_BEFORE=""
 STATE_AFTER=""
 REGISTRY_READ_CALL="credibleRegistry()(address)"
@@ -38,8 +41,10 @@ Usage:
 Target options:
   --target-address ADDRESS                         Existing target on the fresh Anvil chain
   --target-deploy-command COMMAND                  Deploy/upgrade; final stdout line is its address
-  --guarded-call SIGNATURE                         e.g. 'deposit(uint256)' 100
+  --guarded-call SIGNATURE                         e.g. 'setName(string)'
+  --guarded-call-arg ARGUMENT                      Repeat for each argument, e.g. 'hello world'
   --state-read-call SIGNATURE                      e.g. 'totalAssets()(uint256)'
+  --state-read-call-arg ARGUMENT                   Repeat for each state-read argument
   --state-before VALUE                             Expected state before/after rejected calls
   --state-after VALUE                              Expected state after successful guarded call
 
@@ -47,6 +52,7 @@ Registry/configuration options:
   --registry-address ADDRESS                       Existing registry on the fresh Anvil chain
   --registry-deploy-command COMMAND                Deploy/configure; final stdout line is its address
   --marker-call SIGNATURE                          Default: markCurrentBlockCredible()
+  --marker-call-arg ARGUMENT                       Repeat for each marker-call argument
   --expected-threshold BLOCKS                      Default in fixture mode: 10
   --registry-read-call SIGNATURE                   Default: credibleRegistry()(address)
   --threshold-read-call SIGNATURE                  Default: failOpenBlockThreshold()(uint256)
@@ -80,8 +86,11 @@ while [[ $# -gt 0 ]]; do
         --registry-address)        need_value "$@"; REGISTRY_ADDRESS="$2"; shift 2 ;;
         --registry-deploy-command) need_value "$@"; REGISTRY_DEPLOY_COMMAND="$2"; shift 2 ;;
         --guarded-call)            need_value "$@"; GUARDED_CALL="$2"; shift 2 ;;
+        --guarded-call-arg)        need_value "$@"; GUARDED_CALL_ARGS+=("$2"); shift 2 ;;
         --marker-call)             need_value "$@"; MARKER_CALL="$2"; shift 2 ;;
+        --marker-call-arg)         need_value "$@"; MARKER_CALL_ARGS+=("$2"); shift 2 ;;
         --state-read-call)         need_value "$@"; STATE_READ_CALL="$2"; shift 2 ;;
+        --state-read-call-arg)     need_value "$@"; STATE_READ_CALL_ARGS+=("$2"); shift 2 ;;
         --state-before)            need_value "$@"; STATE_BEFORE="$2"; shift 2 ;;
         --state-after)             need_value "$@"; STATE_AFTER="$2"; shift 2 ;;
         --registry-read-call)      need_value "$@"; REGISTRY_READ_CALL="$2"; shift 2 ;;
@@ -118,7 +127,8 @@ if [[ -n "$TARGET_ADDRESS" || -n "$TARGET_DEPLOY_COMMAND" ]]; then
     [[ -n "$REGISTRY_ADDRESS" || -n "$REGISTRY_DEPLOY_COMMAND" ]] ||
         die "target mode requires --registry-address or --registry-deploy-command"
 else
-    [[ -z "$GUARDED_CALL$STATE_READ_CALL$STATE_BEFORE$STATE_AFTER$REGISTRY_ADDRESS$REGISTRY_DEPLOY_COMMAND$CONFIG_ASSERT_COMMAND" ]] ||
+    [[ -z "$GUARDED_CALL$STATE_READ_CALL$STATE_BEFORE$STATE_AFTER$REGISTRY_ADDRESS$REGISTRY_DEPLOY_COMMAND$CONFIG_ASSERT_COMMAND" &&
+        ${#GUARDED_CALL_ARGS[@]} -eq 0 && ${#STATE_READ_CALL_ARGS[@]} -eq 0 ]] ||
         die "target-specific options require --target-address or --target-deploy-command"
     GUARDED_CALL="bump()"
     STATE_READ_CALL="count()(uint256)"
@@ -162,10 +172,13 @@ rpc() { cast rpc --rpc-url "$RPC_URL" "$@" >/dev/null; }
 mine_blocks() { rpc anvil_mine "$(cast to-hex "$1")"; }
 receipt_status() { cast receipt --rpc-url "$RPC_URL" --async "$1" --json 2>/dev/null | jq -r '.status'; }
 receipt_block() { cast to-dec "$(cast receipt --rpc-url "$RPC_URL" --async "$1" --json | jq -r '.blockNumber')"; }
-read_state() { cast call --rpc-url "$RPC_URL" "$TARGET_ADDRESS" $STATE_READ_CALL; }
+read_state() {
+    cast call --rpc-url "$RPC_URL" "$TARGET_ADDRESS" "$STATE_READ_CALL" "${STATE_READ_CALL_ARGS[@]}"
+}
 send_async() {
-    local key="$1" address="$2" call="$3"
-    cast send --rpc-url "$RPC_URL" --private-key "$key" --gas-limit "$GAS_LIMIT" --async "$address" $call
+    local key="$1" address="$2" call="$3"; shift 3
+    cast send --rpc-url "$RPC_URL" --private-key "$key" --gas-limit "$GAS_LIMIT" \
+        --async "$address" "$call" "$@"
 }
 reset_state() {
     rpc evm_revert "$SNAPSHOT"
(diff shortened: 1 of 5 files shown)


examples/credible-block/script/test-credible-upgrades.sh
```diff
@@ -1,226 +1,289 @@
...
-counter()        { cast call --rpc-url "$RPC_URL" "$GUARD" "count()(uint256)"; }
-
-# Submit a tx WITHOUT waiting for it to be mined; echoes the tx hash.
+receipt_block() { cast to-dec "$(cast receipt --rpc-url "$RPC_URL" --async "$1" --json | jq -r '.blockNumber')"; }
+read_state() { cast call --rpc-url "$RPC_URL" "$TARGET_ADDRESS" $STATE_READ_CALL; }
 send_async() {
-    local key="$1"; shift
-    cast send --rpc-url "$RPC_URL" --private-key "$key" --gas-limit "$GAS_LIMIT" --async "$@"
+    local key="$1" address="$2" call="$3"
+    cast send --rpc-url "$RPC_URL" --private-key "$key" --gas-limit "$GAS_LIMIT" --async "$address" $call
```
mateo-mro: These unquoted expansions apply word and pathname expansion to the call text after option parsing. A value such as `setName(string) "hello world"` reaches cast as three arguments, so target mode cannot handle valid string arguments even when the user quotes the whole option. The same pattern appears in `read_state` and the static call. Please represent the signature and arguments as an explicit array and add an argument bearing regression case.

makemake-kbo: Fixed in f076b03. Call signatures and arguments are now stored separately, with repeated `--guarded-call-arg`, `--state-read-call-arg`, and `--marker-call-arg` options feeding quoted Bash arrays into every `cast` invocation. I also added a regression case with the string argument `hello world` and updated the README examples.

examples/credible-block/script/test-credible-upgrades.sh
```diff
@@ -1,226 +1,302 @@
...
+        echo "  ${GREEN}PASS${NC} contract-specific configuration assertion"
+        PASS_COUNT=$((PASS_COUNT + 1))
+    else
+        die "contract-specific configuration assertion failed"
+    fi
+else
+    check "target uses expected registry" \
+        "$(cast call --rpc-url "$RPC_URL" "$TARGET_ADDRESS" "$REGISTRY_READ_CALL")" "$REGISTRY_ADDRESS"
+    check "target uses expected fail-open threshold" \
+        "$(cast call --rpc-url "$RPC_URL" "$TARGET_ADDRESS" "$THRESHOLD_READ_CALL")" "$FAIL_OPEN_THRESHOLD"
```
lean-apple: These are raw string comparisons against `cast` output, but `cast` annotates large numbers: `cast abi-decode "f()(uint256)" 0x…f4240` prints `1000000 [1e6]`, not `1000000`. That means the README's own example (`--state-after 1000000` with `totalAssets()(uint256)`) can never pass, and any threshold or state value ≥ 100000 breaks the same way. Please normalize before comparing (strip the ` [...]` suffix or take the first field). The same applies to the state comparisons via `read_state` (initial-state check and the case checks below).

examples/credible-block/script/test-credible-upgrades.sh
```diff
@@ -1,226 +1,302 @@
...
 done
-
-if ! cast block-number --rpc-url "$RPC_URL" >/dev/null 2>&1; then
-    echo "${RED}ERROR${NC} anvil did not become ready on $RPC_URL within the timeout." >&2
-    exit 1
+cast block-number --rpc-url "$RPC_URL" >/dev/null 2>&1 ||
+    die "Anvil did not become ready on $RPC_URL"
+
+if [[ -n "$REGISTRY_DEPLOY_COMMAND" ]]; then
+    REGISTRY_ADDRESS=$(run_address_command "registry deployment/configuration" "$REGISTRY_DEPLOY_COMMAND")
```
lean-apple, line 229: This registry deployment command runs before `EXPECTED_THRESHOLD` is exported (line 221), yet both the README and `--help` state that deploy commands receive `EXPECTED_THRESHOLD`. A registry command that reads it to configure the threshold silently gets an empty value. Move the export above this invocation, or correct the docs.

examples/credible-block/script/test-credible-upgrades.sh
```diff
@@ -1,226 +1,302 @@
...
+section "Verifying target configuration"
+if [[ -n "$CONFIG_ASSERT_COMMAND" ]]; then
+    if bash -c "$CONFIG_ASSERT_COMMAND"; then
+        echo "  ${GREEN}PASS${NC} contract-specific configuration assertion"
+        PASS_COUNT=$((PASS_COUNT + 1))
+    else
+        die "contract-specific configuration assertion failed"
+    fi
+else
+    check "target uses expected registry" \
```
lean-apple, line 264: The README says the runner "fails before behavioral cases when the assertion fails", but that is only true for the `--config-assert-command` branch above (which `die`s). These default getter checks go through `check`, so the run continues on mismatch; if the target's real threshold differs from `--expected-threshold`, case 3 mines the wrong gap and produces confusing cascade failures. Either `die` when this section records a failure, or read the threshold back from the chain for case 3 (as the previous version of the script did).

examples/credible-block/script/test-credible-upgrades.sh
```diff
@@ -1,226 +1,302 @@
...
+    echo "configuration valid (mode=$($CUSTOM_MODE && echo target || echo fixture), threshold=$FAIL_OPEN_THRESHOLD)"
+    exit 0
+fi
 
-GAS_LIMIT=2000000 # explicit so `cast send` skips eth_estimateGas (which would fail on reverting txs)
+for command in anvil cast forge jq; do
+    command -v "$command" >/dev/null || die "required command not found on PATH: $command"
+done
 
+export FOUNDRY_PROFILE=credible-block
```
lean-apple: This export happens before user-supplied deploy commands run, so a `--target-deploy-command` invoking `forge script` in the caller's own repo inherits a `credible-block` profile that its `foundry.toml` almost certainly lacks, and forge rejects unknown profiles. Scope the profile to the two fixture `forge create` calls instead, or document that deploy commands may need to override `FOUNDRY_PROFILE`.

examples/credible-block/script/test-credible-upgrades.sh
makemake-kbo: Fixed in 3f61549. Numeric `cast` outputs are normalized before threshold and state comparisons, so annotated values such as `100000 [1e5]` compare correctly.

examples/credible-block/script/test-credible-upgrades.sh
makemake-kbo: Fixed in 3f61549. `EXPECTED_THRESHOLD` is now exported before either user-supplied deployment command runs.

examples/credible-block/script/test-credible-upgrades.sh
makemake-kbo: Fixed in 3f61549. Default getter mismatches now abort immediately after configuration verification, before initial-state and behavioral cases.

examples/credible-block/script/test-credible-upgrades.sh
makemake-kbo: Fixed in 3f61549. `FOUNDRY_PROFILE=credible-block` is scoped to the two built-in fixture `forge create` calls, and the README clarifies that user deployment commands retain their caller-provided profile.

examples/credible-block/script/test-credible-upgrades.sh
```diff
@@ -1,226 +1,311 @@
...
-# Submit a tx WITHOUT waiting for it to be mined; echoes the tx hash.
+receipt_block() { cast to-dec "$(cast receipt --rpc-url "$RPC_URL" --async "$1" --json | jq -r '.blockNumber')"; }
+normalize_cast_value() {
+    local value="$1"
+    printf '%s' "${value%% \[*}"
+}
+read_state() {
+    local value
+    value=$(cast call --rpc-url "$RPC_URL" "$TARGET_ADDRESS" "$STATE_READ_CALL" "${STATE_READ_CALL_ARGS[@]}")
+    normalize_cast_value "$value"
```
mateo-mro, line 187: `cast call` prints ABI strings with quotes, so the documented string state example compares `hello world` with `"hello world"` and fails even when the contract state is correct. The argument regression test only parses options and cannot catch this. Could the expected value use the exact `cast` representation, with a live regression case for string state?

mateo-mro (dismissed):

examples/credible-block/script/test-credible-upgrades.sh
makemake-kbo: Fixed in d89cc98. String expectations now use the exact quoted \`cast call\` representation, the README documents it, and the new live \`GuardedName\` target-mode regression exercises the spaced \`hello world\` argument and quoted string state through all three scenarios (14/14 checks passed).

examples/credible-block/script/test-credible-upgrades.sh
```diff
@@ -1,226 +1,311 @@
...
+    command -v "$command" >/dev/null || die "required command not found on PATH: $command"
+done
 
+RPC_URL="http://127.0.0.1:${RPC_PORT}"
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 REPO_ROOT="$(cd "$SCRIPT_DIR/../../.." && pwd)"
+MARKER_ADDRESS=$(cast wallet address "$MARKER_KEY")
+GUARDED_ADDRESS=$(cast wallet address "$GUARDED_KEY")
+export RPC_URL REPO_ROOT ADMIN_KEY MARKER_KEY GUARDED_KEY MARKER_ADDRESS GUARDED_ADDRESS
+export EXPECTED_THRESHOLD="$FAIL_OPEN_THRESHOLD"
```
lean-apple, line 155: **[P2] Export the target before the registry command**

`TARGET_ADDRESS` has been parsed here but is not exported before the registry deployment/configuration command runs. Consequently, `--target-address … --registry-deploy-command '…$TARGET_ADDRESS…'` sees an empty variable even though the README and help text promise it to deployment commands. Please normalize and export a supplied target address before executing the first user command, and add an environment-passing regression case.

examples/credible-block/script/test-credible-upgrades.sh
```diff
@@ -1,226 +1,311 @@
...
+mine_blocks() { rpc anvil_mine "$(cast to-hex "$1")"; }
 receipt_status() { cast receipt --rpc-url "$RPC_URL" --async "$1" --json 2>/dev/null | jq -r '.status'; }
-receipt_block()  { cast to-dec "$(cast receipt --rpc-url "$RPC_URL" --async "$1" --json | jq -r '.blockNumber')"; }
-counter()        { cast call --rpc-url "$RPC_URL" "$GUARD" "count()(uint256)"; }
-
-# Submit a tx WITHOUT waiting for it to be mined; echoes the tx hash.
+receipt_block() { cast to-dec "$(cast receipt --rpc-url "$RPC_URL" --async "$1" --json | jq -r '.blockNumber')"; }
+normalize_cast_value() {
+    local value="$1"
+    printf '%s' "${value%% \[*}"
```
lean-apple: **[P2] Anchor numeric annotation normalization**

This removes everything from the first ` [` anywhere in the value. A valid ABI string such as `"hello [world]"` is therefore normalized to `"hello`, breaking the newly supported string-state comparison. Please strip the annotation only when the complete output matches `cast`’s numeric annotated form and otherwise preserve the value unchanged. A live bracket-containing string case would cover this.

examples/credible-block/script/test-credible-upgrades.sh
```diff
@@ -1,226 +1,311 @@
...
+need_value() { [[ $# -ge 2 && -n "$2" ]] || die "$1 requires a value"; }
+
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+        --target-address)          need_value "$@"; TARGET_ADDRESS="$2"; shift 2 ;;
+        --target-deploy-command)   need_value "$@"; TARGET_DEPLOY_COMMAND="$2"; shift 2 ;;
+        --registry-address)        need_value "$@"; REGISTRY_ADDRESS="$2"; shift 2 ;;
+        --registry-deploy-command) need_value "$@"; REGISTRY_DEPLOY_COMMAND="$2"; shift 2 ;;
+        --guarded-call)            need_value "$@"; GUARDED_CALL="$2"; shift 2 ;;
+        --guarded-call-arg)        need_value "$@"; GUARDED_CALL_ARGS+=("$2"); shift 2 ;;
```
lean-apple: **[P2] Permit empty ABI arguments**

The shared `need_value` check requires `-n "$2"`, so `--guarded-call-arg ""`, `--marker-call-arg ""`, and `--state-read-call-arg ""` are rejected even though an empty string is a valid ABI argument. Please distinguish a missing option operand from a present-but-empty call argument, and add an empty-string target-mode regression.

lean-apple (commented):
​

examples/credible-block/script/test-credible-upgrades.sh
mateo-mro: This normalization runs for every ABI return, so a valid string such as `"hello [world]"` is truncated at the bracket even though the annotation stripping is only needed for numeric output. Could we restrict this to integer return signatures and keep string values untouched, with a bracket bearing string regression?

examples/credible-block/script/test-credible-upgrades.sh
makemake-kbo: Fixed in d68fd52. A supplied target address is checksum-normalized and exported before the registry deployment/configuration command runs, so TARGET_ADDRESS is available to the first user command.

examples/credible-block/script/test-credible-upgrades.sh
makemake-kbo: Fixed in d68fd52. Repeated ABI argument options now distinguish a missing operand from a present empty string. Argument parsing and a live empty-string target run both pass.

mateo-mro (approved):

PR #102, opened 2026-08-03
fix(vault): verify ERC-4626 operation state effects

makemake-kbo:
## Summary
- verify receiver/owner share deltas, total supply, and underlying movement for deposit/mint/withdraw/redeem
- remove the unbounded generic preview deviation; each concrete adapter now pins its supported exact bound
- add assertion-executing honest and malicious state-effect regressions

## Validation
- `pcl test --match-contract ERC4626PreviewAssertionTest` (7 passing)
- Curve and Spark profiles build offline

Linear: ENG-4234

### DIFF
--- a/src/protection/vault/ERC4626PreviewAssertion.sol
+++ b/src/protection/vault/ERC4626PreviewAssertion.sol
@@ -23,12 +23,9 @@ import {ERC4626BaseAssertion} from "./ERC4626BaseAssertion.sol";
 ///       previewWithdraw rounds UP   (returns more shares   -> favors vault)
 ///       previewRedeem   rounds DOWN (returns fewer assets   -> favors vault)
 ///
-/// @dev Uses V2 `registerFnCallTrigger` + `ph.context()` for call-scoped triggers,
-///      `ph.callinputAt()` to read call arguments, and `ph.callOutputAt()` to read the
-///      actual return value — replacing the totalSupply/totalAssets delta inference from V1.
-///      ERC-4626 specifies the conservative direction of previews, but it does not set a generic
-///      maximum distance from the state-changing result. Concrete vaults may override
-///      `_maxPreviewDeviation()` when their implementation proves a tighter bound.
+/// @dev In addition to return values, every operation proves the corresponding receiver/owner
+///      share delta, total-supply delta, and underlying-token movement. ERC-4626 does not define a
+///      universal preview-distance bound, so every concrete supported adapter must provide one.
 abstract contract ERC4626PreviewAssertion is ERC4626BaseAssertion {
     /// @notice Register the default trigger set for preview-consistency invariants.
     /// @dev Each ERC-4626 operation gets its own assertion function via registerFnCallTrigger.
@@ -40,11 +37,8 @@ abstract contract ERC4626PreviewAssertion is ERC4626BaseAssertion {
     }
 
     /// @notice Maximum acceptable deviation between a preview result and the actual result.
-    /// @dev Defaults to no generic distance cap. Override only when the concrete vault's
-    ///      implementation proves a smaller maximum.
-    function _maxPreviewDeviation() internal view virtual returns (uint256) {
-        return type(uint256).max;
-    }
+    /// @dev Must be derived from the concrete vault implementation; no generic default is sound.
+    function _maxPreviewDeviation() internal view virtual returns (uint256);
 
     // ---------------------------------------------------------------
     //  deposit: previewDeposit(assets) <= actualSharesMinted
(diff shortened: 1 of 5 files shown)


test/protection/vault/ERC4626PreviewAssertion.t.sol
```diff
@@ -0,0 +1,192 @@
...
+        _arm(ERC4626PreviewAssertion.assertDepositPreview.selector);
+        vm.expectRevert(bytes("ERC4626: deposit receiver shares mismatch"));
+        vault.deposit(100 ether, receiver);
+    }
+
+    function testMissingAssetEffectTrips() public {
+        vault.setFaults(false, true);
+        _arm(ERC4626PreviewAssertion.assertDepositPreview.selector);
+        vm.expectRevert(bytes("ERC4626: deposit asset payment mismatch"));
+        vault.deposit(100 ether, receiver);
```
mateo-mro, line 232: The adversarial cases only exercise `deposit`, while `mint`, `withdraw`, and `redeem` each have separate return decoding and balance logic. A swapped owner or receiver read, or a wrong payout amount in any of those branches, would still pass this suite. Could we add the same missing share and missing asset cases for the other three operations so the regression coverage matches ENG 4234?

makemake-kbo: Fixed in 13fe43f. Added missing-share and missing-asset adversarial cases for mint, withdraw, and redeem, including their operation-specific receiver/owner and payout checks. The full ERC4626 preview assertion test file passes under PCL.

src/protection/vault/ERC4626PreviewAssertion.sol
```diff
@@ -167,4 +238,48 @@ abstract contract ERC4626PreviewAssertion is ERC4626BaseAssertion {
...
+    function _receiver(bytes memory input, PhEvm.TriggerContext memory ctx) internal view returns (address) {
+        return input.length >= 68 ? _addressArg(input, 1) : _triggerCaller(ctx);
+    }
+
+    function _withdrawAccounts(bytes memory input, PhEvm.TriggerContext memory ctx)
+        internal
+        view
+        returns (address receiver, address owner)
+    {
+        address caller = _triggerCaller(ctx);
```
mateo-mro, line 222: Canonical withdraw and redeem calldata already contain both addresses, but this scans every matching call before checking the input length. A transaction with many exits therefore makes the assertion work quadratic and can hit the execution budget. Could we decode and return immediately when the full arguments are present, then scan only for the shorter overloads?

src/protection/vault/ERC4626PreviewAssertion.sol
```diff
@@ -130,19 +183,37 @@ abstract contract ERC4626PreviewAssertion is ERC4626BaseAssertion {
...
+            _assetBalanceAt(_assetCustodyAccount(), pre),
+            _assetBalanceAt(_assetCustodyAccount(), post),
+            actualAssets,
+            "ERC4626: redeem vault assets mismatch"
+        );
+        _requireIncrease(
+            _assetBalanceAt(receiver, pre),
+            _assetBalanceAt(receiver, post),
+            actualAssets,
+            "ERC4626: redeem receiver assets mismatch"
```
mateo-mro: These custody and receiver checks assume different accounts. ERC 4626 does not forbid `receiver` from being the vault, and the Llama adapter can likewise receive at its controller. A self transfer succeeds with no net asset balance change, so this rejects a valid withdrawal or redemption. Could overlapping accounts be handled as one net delta and covered by a regression?

test/protection/vault/ERC4626PreviewAssertion.t.sol
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    function testRedeemMissingAssetEffectTrips() public {
+        vault.deposit(100 ether, address(this));
+        vault.setFaults(false, true);
+        _arm(ERC4626PreviewAssertion.assertRedeemPreview.selector);
+        vm.expectRevert(bytes("ERC4626: redeem vault assets mismatch"));
+        vault.redeem(40 ether, receiver, address(this));
+    }
+}
```
mateo-mro, line 351: The shared mock covers only the zero deviation fixture. None of these tests instantiate LlamaLend, Spark, or MetaMorpho after their new bounds and custody behavior, and the PR validation lists builds only for the adapter profiles. ENG 4234 asks for honest and incorrect state effect regressions for every active adapter. Could we add focused PCL cases for each adapter, including the Llama controller custody path?

src/protection/vault/ERC4626PreviewAssertion.sol
```diff
@@ -55,20 +57,35 @@ abstract contract ERC4626PreviewAssertion is ERC4626BaseAssertion {
...
         require(previewShares <= actualShares, "ERC4626: previewDeposit > actual shares");
         require(actualShares - previewShares <= _maxPreviewDeviation(), "ERC4626: deposit preview deviates from actual");
+        _requireIncrease(
+            _shareBalanceAt(receiver, pre),
+            _shareBalanceAt(receiver, post),
+            actualShares,
+            "ERC4626: deposit receiver shares mismatch"
+        );
+        _requireIncrease(_totalSupplyAt(pre), _totalSupplyAt(post), actualShares, "ERC4626: deposit supply mismatch");
+        _requireIncrease(
```
lean-apple: **[P1] Verify the caller’s asset debit**

The custody increase alone does not prove that the caller paid exactly `assets`. A faulty deposit could pull `2 * assets` from the caller, divert half elsewhere, mint/return the expected shares, and leave custody with the required net increase; every current check would pass. Please also verify the triggering caller’s asset balance decreases by `assets` for deposit and `actualAssets` for mint, with appropriate handling when payer and custody are the same account, and add an overcharge/wrong-payer regression.

src/protection/vault/ERC4626PreviewAssertion.sol
lean-apple, line 222: **[P2] Avoid the unbounded caller scan for canonical exits**

Canonical `withdraw`/`redeem` calldata already supplies both receiver and owner, but this unconditionally calls `_triggerCaller`. That materializes and scans every matching call in the transaction for each assertion invocation, so a valid batch of exits can exceed the fixed assertion budget. Resolve the caller only when an overload actually omits an account, and use a bounded call lookup for those overloads.

lean-apple (commented):
​

src/protection/vault/ERC4626PreviewAssertion.sol
```diff
@@ -167,4 +208,102 @@ abstract contract ERC4626PreviewAssertion is ERC4626BaseAssertion {
...
+        address caller = _triggerCaller(ctx);
+        receiver = input.length >= 68 ? _addressArg(input, 1) : caller;
+        owner = caller;
+    }
+
+    function _triggerCaller(PhEvm.TriggerContext memory ctx) internal view returns (address) {
+        PhEvm.CallFilter memory filter = PhEvm.CallFilter({
+            callType: 0, minDepth: 0, maxDepth: type(uint32).max, topLevelOnly: false, successOnly: true
+        });
+        PhEvm.TriggerCall[] memory calls = ph.matchingCalls(vault, ctx.selector, filter, 8);
```
mateo-mro, line 231: Capping this lookup at eight makes the ninth valid deposit or mint in one transaction revert because its call id can never be found. That turns the performance bound into an undocumented protocol restriction. Could we either resolve the caller without a global prefix scan or explicitly enforce and document a supported call count at registration?

mateo-mro (approved):

PR #103, opened 2026-08-03
fix(safe): close transaction-shape and guard validation gaps [ENG-4235]

makemake-kbo:
## Summary
- make the shared Safe protection contracts the single maintained source while preserving example imports
- enforce CALL MultiSend outer-value and actual allowance-owner semantics, and reject transient cumulative allowance grants above policy caps
- align deployment registry validation with runtime 50k-gas, exact-return, canonical-bool, and future-block rules
- document tested module-set limits and add regressions for CALL batches, transient approvals, malformed/slow registries, and future credible blocks

## Validation
- `FOUNDRY_PROFILE=safe pcl test` (2 passed)
- targeted Safe PCL regressions for transient grants, CALL owner semantics, and outer value (3 passed)
- `FOUNDRY_PROFILE=safe-guard forge test --offline --match-path test/protection/safe/integration/CredibleSafeGuardScripts.t.sol` (18 passed)
- `forge fmt --check`

Linear: ENG-4235


src/protection/safe/SafeTxShapeHelpers.sol
```diff
@@ -293,11 +293,52 @@ abstract contract SafeTxShapeHelpers is Assertion {
         while (offset < transactionsLength) {
             (Action memory innerAction, uint256 nextOffset) =
                 _readMultiSendAction(action, transactionsOffset, transactionsLength, offset);
-            _validateInnerApprovalPolicy(innerAction);
+            if (
+                innerAction.operation == OPERATION_CALL && innerAction.dataLength == 68
+                    && _selectorAt(innerAction.data, innerAction.dataOffset) == INCREASE_ALLOWANCE_SELECTOR
+            ) {
+                address spender = _readAbiAddress(innerAction.data, innerAction.dataOffset + 4);
+                uint256 cumulativeIncrease = _cumulativeIncreaseAllowance(
```
mateo-mro: Every `increaseAllowance` entry rescans the batch from the first action through its own offset, so a batch containing many grants now does quadratic decoding work. Since `maxActions` has no global upper bound, an otherwise allowed batch can exhaust the assertion gas budget. Could this accumulate grants in one pass, or enforce and document a measured batch bound?

makemake-kbo: Fixed in 69a56c0. Batch grants are accumulated in one decode pass by approval-policy index, each pre-state allowance is read once, and batch policies are globally capped at four actions with the approval assertion enforcing the configured cap independently. The four-grant regression uses 264,802 assertion gas under PCL's 300,000-gas local ceiling; the full Safe shape suite passes.

src/protection/safe/SafeTxShapeHelpers.sol
```diff
@@ -289,11 +292,52 @@ abstract contract SafeTxShapeHelpers is Assertion {
...
+                        _readAllowanceFromPreState(innerAction.caller, innerAction.target, spender);
+                    allowanceLoaded[policyIndex] = true;
+                }
+
+                uint256 cumulativeIncrease = cumulativeIncreases[policyIndex];
+                cumulativeIncrease = addedValue > type(uint256).max - cumulativeIncrease
+                    ? type(uint256).max
+                    : cumulativeIncrease + addedValue;
+                cumulativeIncreases[policyIndex] = cumulativeIncrease;
+                _validateAllowancePeak(innerAction.target, spender, initialAllowances[policyIndex], cumulativeIncrease);
```
mateo-mro: The accumulator only includes `increaseAllowance`. A batch can call `approve(spender, 100)` and then `increaseAllowance(spender, 100)` with a cap of 100: the approval passes independently, the increase is checked against the pretransaction allowance of zero, and the spender receives 200. This is the transient grant bypass ENG 4235 is meant to close. Could the single pass model every positive ERC20 grant for each policy, including direct approvals and grants consumed or reset later in the batch?

mateo-mro (changes requested):

src/protection/safe/SafeTxShapeHelpers.sol
```diff
@@ -289,11 +292,52 @@ abstract contract SafeTxShapeHelpers is Assertion {
...
         uint256 offset;
+        uint256 actionCount;
         while (offset < transactionsLength) {
             (Action memory innerAction, uint256 nextOffset) =
                 _readMultiSendAction(action, transactionsOffset, transactionsLength, offset);
-            _validateInnerApprovalPolicy(innerAction);
+            ++actionCount;
+            if (actionCount > batchPolicy.maxActions) revert SafeTxShapeBatchTooManyActions(batchPolicy.maxActions);
+            if (
+                innerAction.operation == OPERATION_CALL && innerAction.dataLength == 68
```
lean-apple, line 318: **[P1] Accumulate mixed approval grant methods**

This only accumulates `increaseAllowance` entries. With both baseline policies capped at 100, a batch containing `approve(spender, 80)` followed by `increaseAllowance(spender, 30)` passes: the approve is checked independently, while the increase calculation starts from the pre-batch allowance of zero and sees only 30, even though the actual transient allowance reaches 110. Please track one ordered allowance state per owner/token/spender across both positive `approve` and `increaseAllowance` actions—or reject mixed grant methods—and add this regression.

lean-apple (commented):
​

src/protection/safe/SafeTxShapeHelpers.sol
makemake-kbo: Fixed in e6ade7b. The batch approval scan now rejects mixing ERC-20 `approve` and positive `increaseAllowance` grant methods, closing the `approve(100) + increaseAllowance(100)` transient-cap bypass. Added regressions for both method orderings; the relevant three transient-grant PCL tests pass.

src/protection/safe/SafeTxShapeHelpers.sol
makemake-kbo: Fixed in e6ade7b using the accepted reject-mixed-methods alternative. A batch cannot combine ERC-20 `approve` with positive `increaseAllowance`; regressions cover `approve(80) -> increaseAllowance(30)` and the reverse ordering. Runtime size is 24,526 bytes, 50 bytes below EIP-170.

src/protection/safe/README.md
```diff
@@ -179,7 +179,7 @@ Approval policy:
...
+For ERC-20 `approve(spender, amount)` the cap binds `amount` directly. For ERC-20 `increaseAllowance(spender, addedValue)` the cap binds the pre-execution allowance plus all positive grants requested for that owner/token/spender in the batch. This is intentionally stricter than checking final state: consuming or reducing an oversized transient allowance later in the transaction does not make the batch valid. The approval assertion decodes the batch once, accumulates each configured allowance by policy index, and reads each initial allowance once, so multiple grants do not trigger repeated prefix rescans. Batch policies cannot configure more than four actions; the four-grant regression executes below PCL's 300,000-gas local assertion ceiling, while the former unbounded configuration could exhaust the budget before reaching a policy decision. For CALL-based executors, the executor is the token owner; for DELEGATECALL-based executors, the Safe is the owner.
```
mateo-mro, line 182: These flags apply to the whole batch, so an approve for token A followed by increaseAllowance for unrelated token B is rejected too. That is a reasonable conservative choice, but ENG 4235 asks us to call out behavior that is stricter than Safe. Could this paragraph explicitly say that mixing these ERC20 grant methods anywhere in one batch is unsupported, even across different tokens or spenders?

mateo-mro (dismissed):

src/protection/safe/README.md
makemake-kbo: Fixed in 711df59. The README now states explicitly that mixing approve and positive increaseAllowance anywhere in a batch is unsupported, including across unrelated tokens, owners, and spenders.

mateo-mro (approved):

PR #105, opened 2026-08-03
feat(aave): harden v3 oracle guard and add v4 protections

makemake-kbo:
## Summary

- harden the Aave v3 Horizon oracle assertion and expand adversarial and boundary coverage
- add Aave v4 oracle-consumption and external-collateral transferability assertions
- add an Aave v4 Hub flow-rate circuit breaker and PT-USDe redemption protection

## Testing

- Solidity changes compile successfully
- full Foundry test execution is currently blocked locally by a Foundry crash during system proxy initialization


examples/aave/test/AaveV3AdversarialResearch.t.sol
```diff
@@ -0,0 +1,626 @@
...
+        function testBackingSameSeizureTripsWhenTransactionAlsoTouchesPool() public {
+            _armBacking();
+            vm.expectRevert(bytes("AaveV3Horizon: reserve backing deficit"));
+            bundle.seizeThenPoolCall(
+                collateralAsset, address(collateralAToken), recipient, 1 ether, pool, address(debtAsset), alice
+            );
+        }
+
+        /// @dev The upstream liability is aToken supply plus scaled accrued treasury; the assertion omits the latter.
+        function testBackingOmittedTreasuryClaimPassesDespiteEconomicShortfall() public {
```
mateo-mro: This test proves the published reserve backing assertion accepts a real liability shortfall, but the pull request leaves that assertion unchanged and still lists it as an Aave example. Shipping a passing test for the bypass does not protect adopters. Could we quarantine the assertion or include the accrued treasury liability before keeping it advertised?

makemake-kbo: Fixed in c962a4a. The backing assertion now converts scaled \`accruedToTreasury\` with Aave's half-up ray multiplication and includes that amount on the liability side. The former false-negative now expects and receives the backing-deficit revert; all five backing-focused PCL tests pass.

examples/aave/src/AaveV3HorizonOracleAssertion.sol
```diff
@@ -3,189 +3,265 @@ pragma solidity ^0.8.13;
...
+    }
 
-        if (selector == IAaveV3LikePool.setUserUseReserveAsCollateral.selector) {
-            (address asset,) = abi.decode(input, (address, bool));
-            _assertSourceStable(oracle, asset, pre, post);
-            _assertPriceBounded(oracle, asset, pre, post, ORACLE_DEVIATION_BPS);
-            return;
-        }
+    function _assertPoolProviderReturns(address expectedOracle) internal view {
+        PhEvm.CallInputs[] memory providerCalls =
```
mateo-mro, line 165: Neither this provider scan nor the consumed price scan requires a matching call to exist. A successful borrow can skip `getPriceOracle` and the configured oracle entirely; both arrays are empty and the assertion passes. That leaves the guard open to the exact price path bypass it is intended to catch. Could mandatory risk operations fail when the expected provider and oracle calls are absent, with a no price borrow regression?

examples/aave/src/AaveV4OracleConsumptionHelpers.sol
```diff
@@ -0,0 +1,80 @@
...
+        });
+        return ph.matchingCalls(target, selector, filter, limit);
+    }
+
+    function _successfulSourceReads(address source, uint256 limit) internal view returns (PhEvm.TriggerCall[] memory) {
+        return _successfulStaticCalls(source, IAaveV4PriceFeed.latestAnswer.selector, limit);
+    }
+
+    function _hasMandatoryPriceOperation(address spoke) internal view returns (bool) {
+        return _matchingCalls(spoke, IAaveV4Spoke.borrow.selector, 1).length != 0
```
mateo-mro, line 82: This mandatory set omits the other risk paths named by the assertion: withdraw, collateral disable, risk premium refresh, and dynamic config refresh. If one of those paths stops reading the oracle, the transaction returns before the implementation and routing checks, so an implementation write and restore can pair with a price bypassing withdrawal without being examined. Could mandatory detection cover every operation that must consume prices, with a regression where one succeeds without a price read?

examples/aave/src/AaveV4OracleConsumptionAssertion.sol
```diff
@@ -0,0 +1,409 @@
...
+        ORACLE = oracle_;
+        EXPECTED_SPOKE_IMPLEMENTATION = expectedSpokeImplementation_;
+        MAX_TRACE_CALLS = maxTraceCalls_;
+
+        for (uint256 i; i < reservePolicies_.length; ++i) {
+            ReservePolicy memory policy = reservePolicies_[i];
+            require(policy.reserveId == i, "AaveV4Oracle: policies not contiguous");
+            require(policy.asset != address(0), "AaveV4Oracle: policy asset zero");
+            require(policy.source != address(0), "AaveV4Oracle: policy source zero");
+            require(policy.deviationBps <= BPS, "AaveV4Oracle: bad tolerance");
```
mateo-mro: Allowing 10,000 bps makes the lower bound zero, so any nonzero price collapse passes. PR 100 rejects this value for the existing active oracle guards for the same reason, but it cannot cover this new contract. Could this require `deviationBps < BPS` and add the boundary case?

examples/aave/README.md
```diff
@@ -16,5 +16,53 @@ FOUNDRY_PROFILE=aave forge build
...
+
+## Aave v4 flow-rate calibration
+
+The ready-to-adopt Core and Prime Hub circuit breakers cover WBTC, USDG, and
+wstETH, the three highest-TVL Aave v4 assets in the 2026-07-30 DefiLlama
+snapshot. Their 24-hour cumulative-flow and 10-second peak-rate limits use a
+20% buffer over the maximum observed values in the preceding 30 days.
+
+See
+[aave-v4-flow-rate-calibration-2026-07-30.md](research/aave-v4-flow-rate-calibration-2026-07-30.md)
```
mateo-mro, line 75: These references are not present in the branch, so all three links render as missing files after merge. Could we add the deployment and research documents or remove the links?

mateo-mro (changes requested):

examples/aave/src/AaveV3HorizonOracleAssertion.sol
makemake-kbo: Fixed in c8f82c9. Mandatory v3 risk operations now require at least one Pool-originated `getPriceOracle` call and at least one mapped Pool-originated `getAssetPrice` call. Added `testBorrowWithoutProviderOrPriceReadFailsClosed`; the focused PCL suite passes 14/14.

examples/aave/src/AaveV4OracleConsumptionHelpers.sol
makemake-kbo: Fixed in c8f82c9. Mandatory v4 detection now covers borrow, withdraw, liquidation, collateral disable, risk-premium refresh, and dynamic-config refresh. Added no-price-read regressions for all six paths while preserving collateral enable as a non-price path; the focused PCL suite passes 35/35.

examples/aave/src/AaveV4OracleConsumptionAssertion.sol
makemake-kbo: Fixed in c8f82c9. Reserve policies now require `deviationBps < BPS`, and `testFullRangeToleranceIsRejected` covers the 10,000 bps boundary.

examples/aave/README.md
makemake-kbo: Fixed in c8f82c9. Added the deployment guide and both linked research/calibration documents, so all three README references resolve in the branch.

examples/aave/test/AaveV3AdversarialResearch.t.sol
mateo-mro: The stored liquidityIndex is not the index used by aToken totalSupply or by a future treasury mint once time has elapsed. Those use the current normalized income, so an untouched reserve can still understate this liability and preserve backing slack. Could we read getReserveNormalizedIncome at the same fork and add a regression where the stored index is stale?

examples/aave/src/AaveV3HorizonOracleAssertion.sol
```diff
@@ -3,189 +3,271 @@ pragma solidity ^0.8.13;
...
         POOL = pool_;
         ADDRESSES_PROVIDER = addressesProvider_;
-        MAX_RESERVES_TO_SCAN = maxReservesToScan_;
-        ORACLE_DEVIATION_BPS = oracleDeviationBps_;
+        MAX_TRACE_CALLS = maxTraceCalls_;
+
+        for (uint256 i; i < assetPolicies_.length; ++i) {
+            AssetPolicy memory policy = assetPolicies_[i];
+            require(policy.asset != address(0), "AaveV3Horizon: policy asset zero");
+            require(policy.deviationBps <= BPS, "AaveV3Horizon: bad asset tolerance");
```
mateo-mro: V4 now rejects 10000 bps because it makes the lower bound zero, but this constructor still accepts it. With that setting, any positive collapse can pass the envelope. Could we require deviationBps to be below BPS here as well and cover the boundary?

examples/aave/src/AaveV3HorizonReserveBackingAssertion.sol
mateo-mro, line 1: The review document demonstrates that a direct seizure without a Pool call executes zero assertions. This trigger therefore does not cover the direct token mutations described above, and a later Pool transaction only discovers the loss after it has committed. Can we quarantine this assertion or arm a surface that is selected by those token only transactions?

examples/aave/src/AaveV4HubFlowRateCircuitBreaker.sol
```diff
@@ -0,0 +1,168 @@
...
+        require(flow.token != address(0) && rate.token == flow.token, "AaveV4Flow: bad outflow context");
+
+        FlowLimits memory limits = _flowLimits(flow.token);
+        require(flow.currentBps <= limits.outflowWindowBps, "AaveV4Flow: 24h outflow limit");
+        require(rate.peakRateBps <= limits.outflowPeakRateBps, "AaveV4Flow: outflow rate limit");
+    }
+
+    function _watchAsset(address token) internal view {
+        watchCumulativeInflow(token, DISPATCH_THRESHOLD_BPS, FLOW_WINDOW, this.assertInflowWithinRateLimits.selector);
+        watchCumulativeOutflow(token, DISPATCH_THRESHOLD_BPS, FLOW_WINDOW, this.assertOutflowWithinRateLimits.selector);
```
mateo-mro, line 79: These rate checks are only reachable through net cumulative flow triggers. If the window already has legitimate net inflow, a fast outflow can leave the net positive, so the outflow trigger never runs and outflowRate is never checked. The reverse applies after prior outflow. That leaves the peak rate limits bypassable when direction changes inside the window. Can this remain unarmed until rate or absolute directional dispatch is available?

examples/aave/research/aave-v4-flow-rate-calibration-2026-07-30.md
```diff
@@ -0,0 +1,72 @@
...
+and comes from the same 10-second buckets.
+
+## Results
+
+| Hub | Asset | Current balance | 30d gross in | 30d gross out | Max 24h net in | In limit | Max 24h net out | Out limit | Peak in rate | In-rate limit | Peak out rate | Out-rate limit |
+| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
+| Core | WBTC | 670.9402 | 129.0485 | 20.8481 | 986.44 bps | 1,184 bps | 91.41 bps | 110 bps | 39.32 bps/s | 48 bps/s | 7.22 bps/s | 9 bps/s |
+| Core | USDG | 30.6733m | 68.6104m | 61.5900m | 4,329.38 bps | 5,196 bps | 5,364.94 bps | 6,438 bps | 438.64 bps/s | 527 bps/s | 128.27 bps/s | 154 bps/s |
+| Core | wstETH | 8,490.4974 | 4,365.9474 | 1,416.9640 | 1,587.63 bps | 1,906 bps | 776.51 bps | 932 bps | 47.57 bps/s | 58 bps/s | 77.62 bps/s | 94 bps/s |
+| Prime | WBTC | 136.7563 | 57.6882 | 47.4818 | 1,750.94 bps | 2,102 bps | 1,972.12 bps | 2,367 bps | 123.91 bps/s | 149 bps/s | 148.27 bps/s | 178 bps/s |
```
mateo-mro: The production limits are described as the observed maximum plus 20 percent, but the checked in chart reports Prime WBTC max out at 2026 bps while this row uses 1972.12 and derives 2367. A 20 percent buffer on the chart value is 2432, so the hardcoded limit has less headroom than stated. Core USDG also differs between the chart and table. Could we reconcile these values and include the calculation script or raw output used for the constants?

examples/aave/src/AaveV3HorizonOracleAssertion.sol
makemake-kbo: Fixed in 1e08b44. The v3 constructor now requires deviationBps < BPS, matching v4, with a 10,000-bps rejection regression.

mateo-mro (dismissed):

examples/aave/src/AaveV3HorizonOracleAssertion.sol
```diff
@@ -3,189 +3,271 @@ pragma solidity ^0.8.13;
...
+        uint256 upperBound = ph.mulDivUp(baselinePrice, BPS + deviationBps, BPS);
+        require(
+            consumedPrice >= lowerBound && consumedPrice <= upperBound, "AaveV3Horizon: consumed oracle price deviated"
+        );
+    }
+
+    function _hasRiskOperation() internal view returns (bool) {
+        return _matchingCalls(POOL, IAaveV3LikePool.borrow.selector, 1).length != 0
+            || _matchingCalls(POOL, IAaveV3LikePool.withdraw.selector, 1).length != 0
+            || _matchingCalls(POOL, IAaveV3LikePool.setUserUseReserveAsCollateral.selector, 1).length != 0
```
mateo-mro, line 267: Every successful collateral toggle is treated as a mandatory price path here. In Aave v3, enabling collateral does not read the oracle; only disabling reaches health factor validation. A normal enable therefore succeeds in the Pool and is then rejected with Pool skipped oracle provider. Could we distinguish the boolean argument and add an honest enable regression, as the v4 guard already does?

examples/aave/src/AaveV3HorizonOracleAssertion.sol
```diff
@@ -3,189 +3,271 @@ pragma solidity ^0.8.13;
...
 
-        if (selector == IAaveV3LikePool.setUserUseReserveAsCollateral.selector) {
-            (address asset,) = abi.decode(input, (address, bool));
-            _assertSourceStable(oracle, asset, pre, post);
-            _assertPriceBounded(oracle, asset, pre, post, ORACLE_DEVIATION_BPS);
-            return;
-        }
+    function _assertPoolProviderReturns(address expectedOracle) internal view {
+        PhEvm.CallInputs[] memory providerCalls =
+            ph.getStaticCallInputs(ADDRESSES_PROVIDER, IAaveV3LikeAddressesProvider.getPriceOracle.selector);
```
mateo-mro, line 166: MAX_TRACE_CALLS is checked only after getStaticCallInputs has already materialized every match. A transaction with enough provider or oracle reads can exhaust the assertion budget before either check runs, so the configured limit does not bound the work. Could these scans use matchingCalls with MAX_TRACE_CALLS plus one, like the source scan below?

mateo-mro (dismissed):

mateo-mro (approved):

